PR #7, opened 2016-01-26
Fix maven-deploy

gilles-duboscq:
We used to guess scm metadata from `hg paths`, this was already not 100% reliable with hg and seems harder in git. We should just let `suite.py` explicitly define the SCM metadata.
Along with the fix to GitConfig._path this should make it possible to use `maven-deploy` on git repos.


gilles-duboscq:
I had to close & reopen to get travis-ci to trigger

PR #8, opened 2016-01-26
Fix usage of git remote

tkrodriguez:
typo that showed up trying to run maven-deploy


dougxc:
Gilles fixed this in a separate push that I just accepted.

PR #11, opened 2016-01-26
remove use of sudo in Travis config

dougxc:


gilles-duboscq:
Interesting error in the 2.6 Travis job. `from mx import DateTime` is really code from `optik_ext.py` that is used by pylint.

dougxc:
Rather than figure out why this is happening, I say we leave the sudo usage in.

gilles-duboscq:
I noticed that in fact this is actually showing that we are not 2.6 compatible: on the non-sudo infrastructure we really get python 2.6 and we have syntax problems.
All the other travis builds that seem to be using 2.6 are actually running on 2.7: the mx script is picking up a 2.7 install which, while not default, is present on the sudo-enabled builders.

dougxc:
Currently, mx is not 2.6 compatible. The few uses of [multi-context-manager with statements](https://docs.python.org/dev/whatsnew/2.7.html#other-language-changes) and single use of OrderedDict can be replaced with 2.6 compatible code. However, the `argparse` module was only [introduced](https://docs.python.org/2.7/whatsnew/2.7.html#pep-389-the-argparse-module-for-parsing-command-lines) in 2.7. A workaround for this is to simply place `argparse.py` on PYTHONPATH.
The bigger question is whether we really need to worry about 2.6 compatibility?

mbrantner:
Oracle Linux 6.5 comes with Pyton 2.6.6. It would be great if that was supported.

dougxc:
I can make mx parse with 2.6 fairly easy by making the changes mentioned above. However, usages of certain 2.7 features will only be caught dynamically. The biggest offender I know of is ["Changed in version 2.7: The positional argument specifiers can be omitted, # so '{} {}' is equivalent to '{0} {1}'."](https://docs.python.org/2/library/string.html#format-string-syntax).

gilles-duboscq:
The `{}` can easily be grepped.
In any case, we should either make it 2.6 compatible or stop testing using 2.6.
Regarding `argparse`, that would be an other advantage of being a python package with a `setup.py`: we could just require `argparse` as a dependency.
One problem there is that there is already a python package called `mx`on pypi...

dougxc:
Grepping for the pattern `"[^"\n]+\{\}` shows 39 hits and `"[^"\n]+\{\}` shows 106 hits in the mx related Python files I have open in Eclipse. These will have to be converted. Unfortunately, I can't see a pylint check to prevent them creeping back in.

dougxc:
I've further investigated the work required to make mx 2.6 compatible and enforce this compatibility on all mx extensions and it's not exactly trivial due to the issue with `{}` in format specifiers mentioned above. One can use the `ast` Python module to parse for string constants containing `{}` but it's hard to know whether they're used as format strings. Also, there's nothing preventing a format string being non-constant. In any case, until there's a more compelling reason to support 2.6, I'm going to disable 2.6 testing.

PR #14, opened 2016-01-27
Fixes for older versions of git

chumer:


gilles-duboscq:
We should still sort the tags.
Since there are anyway other problems with the Git version of `release_version_from_tags` and we need to remove `--sort=refname` for compatibility with older git versions i'll merge that and create an issue about the further required fixes.

gilles-duboscq:
Created #15

Issue #15, opened 2016-01-27
Broken GitConfig.release_version_from_tags


gilles-duboscq:
`GitConfig.release_version_from_tags` does not sort tags properly and does not do "+1 -SNAPSHOT" when not exactly on a tag

PR #16, opened 2016-01-27
Fix release_version_from_tags to match expected behaviour

gilles-duboscq:
* Use `git log` instead of `git tags` since we only want the tags that are ancestors of the current commit.
* Sort versions.
* Only call to git to get the revision number of the most recent version.
* Use `minor + 1` when not exactly at the tag.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -3219,28 +3219,34 @@ class GitConfig(VC):
             else:
                 return None
 
-    def _prefix(self, prefix):
-        if not prefix.endswith('-'):
-            prefix = '{}-'.format(prefix)
-        return prefix
-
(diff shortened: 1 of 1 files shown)


gilles-duboscq:
Fixes #15

mx.py
```diff
@@ -3219,28 +3219,34 @@ def parent(self, vcdir, abortOnError=True):
             else:
                 return None
 
-    def _prefix(self, prefix):
-        if not prefix.endswith('-'):
-            prefix = '{}-'.format(prefix)
-        return prefix
-
     def _tags(self, vcdir, prefix, abortOnError=True):
         # git -C . tag -l prefix-*
```
dougxc: comment should be in terms of log command

PR #19, opened 2016-02-01
fixed git pull

cosminbasca:
Changed the behavior of GitConfig.pull. 
Original behavior: when update=True, do a git pull which works well when the refspec is not a commit hash.
New behavior: when update=True and rev not None, delegate to GitConfig.udpate (with force=True, discard local changes). Should this behavior be different? 

Bonus, updated GitConfig.update to suppress the "detached HEAD" warning messages when not in verbose mode. 

Bumped version to 5.6.15

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -3459,7 +3459,6 @@ class GitConfig(VC):
                 if abortOnError:
                     abort('fetch returned ' + str(rc))
                 return False
-        
 
     def can_push(self, vcdir, strict=True, abortOnError=True):
         """
--- a/mx.py
+++ b/mx.py
@@ -12265,7 +12265,7 @@ def main():
         # no need to show the stack trace when the user presses CTRL-C
         abort(1)
 
-version = VersionSpec("5.6.14")
+version = VersionSpec("5.6.15")
 
 currentUmask = None
 
--- a/mx.py
+++ b/mx.py
@@ -3440,13 +3440,17 @@ class GitConfig(VC):
         rc = self._fetch(vcdir, abortOnError)
         if rc == 0 and update:
             active_branch = self._active_branch(vcdir, abortOnError)
-            refspec = '{0}:{1}'.format(rev if rev else 'HEAD', active_branch)
-            cmd = ['git', 'pull', 'origin', refspec]
-            self._log_pull(vcdir, rev)
-            out = OutputCapture()
-            rc = self.run(cmd, nonZeroIsFatal=abortOnError, cwd=vcdir, out=out)
-            logvv(out.data)
-            return rc == 0
+            if rev:
+                return self.update(vcdir, rev=rev, mayPull=False, clean=True, abortOnError=abortOnError)
+            else:
+                # refspec = '{0}:{1}'.format(rev if rev else 'HEAD', active_branch)
+                refspec = 'HEAD:{0}'.format(active_branch)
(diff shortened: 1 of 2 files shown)


mx.py
```diff
@@ -3445,13 +3445,17 @@ def pull(self, vcdir, rev=None, update=False, abortOnError=True):
...
-            out = OutputCapture()
-            rc = self.run(cmd, nonZeroIsFatal=abortOnError, cwd=vcdir, out=out)
-            logvv(out.data)
-            return rc == 0
+            if rev:
+                return self.update(vcdir, rev=rev, mayPull=False, clean=True, abortOnError=abortOnError)
+            else:
+                # refspec = '{0}:{1}'.format(rev if rev else 'HEAD', active_branch)
+                refspec = 'HEAD:{0}'.format(active_branch)
+                cmd = ['git', 'pull', 'origin', refspec]
```
dougxc: @gilles-duboscq stated that it seems wrong to execute a `pull` after having executed `fetch` given that `pull` also does a `fetch`. I'd like to get his input of this change before merging it.

chumer: I see but. I desperately need this change as graal-vm is broken due to this.
Doing a pull after a fetch is definitely not what we wnat if a revision is specified.

cosminbasca: if I understand correctly, the fetch should only occur on the code branch where we do an update. If we go with the pull we don't do a fetch. Is this correct?

mx.py
```diff
@@ -3445,13 +3445,17 @@ def pull(self, vcdir, rev=None, update=False, abortOnError=True):
...
-            cmd = ['git', 'pull', 'origin', refspec]
-            self._log_pull(vcdir, rev)
-            out = OutputCapture()
-            rc = self.run(cmd, nonZeroIsFatal=abortOnError, cwd=vcdir, out=out)
-            logvv(out.data)
-            return rc == 0
+            if rev:
+                return self.update(vcdir, rev=rev, mayPull=False, clean=True, abortOnError=abortOnError)
+            else:
+                # refspec = '{0}:{1}'.format(rev if rev else 'HEAD', active_branch)
```
dougxc: Is there a good reason to leave this commented code in?

cosminbasca: no, the comment can be removed

dougxc: Ok, can you please remove it.

mx.py
dougxc: Sounds like the right thing to me.

chumer:
I am merging because it was broken anyway. But @gilles-duboscq can you have another look? Thanks.

mx.py
```diff
@@ -3442,20 +3442,23 @@ def pull(self, vcdir, rev=None, update=False, abortOnError=True):
...
             logvv(out.data)
             return rc == 0
         else:
-            if abortOnError:
-                abort('fetch returned ' + str(rc))
-            return False
+            rc = self._fetch(vcdir, abortOnError)
+            if rc == 0:
+                if rev:
+                    return self.update(vcdir, rev=rev, mayPull=False, clean=True, abortOnError=abortOnError)
```
gilles-duboscq: I'm not sure about the `clean=True` that looks dangerous.
In the `hg` implementation we would not clean and rather abort if we can not update cleanly.

Also we should only update if `update` is true

Issue #20, opened 2016-02-02
Suite specific global options do not allow spaces in arguments.


zapster:
Suite specific global options do not allow spaces in arguments, not even if they are quoted.

Without `=`:
```
$ mx --vmprefix "echo -n" vm -version
Suite specific global options must use name=value format: --vmprefix=echo -n
```
With `=` and space:
```
$ mx --vmprefix="echo -n" vm -version
Traceback (most recent call last):
  File "/.../mx.py", line 12275, in <module>
    main()
  File "/.../mx.py", line 12197, in main
    commandAndArgs = _argParser._parse_cmd_line(_opts, firstParse=False)
  File "/.../mx.py", line 6505, in _parse_cmd_line
    abort('Suite specific global options must use name=value format: {0}={1}'.format(self.unknown[-1], self.initialCommandAndArgs[0]))
IndexError: list index out of range
```
With `=` without space:
```
$ mx --vmprefix="echo" vm -version
/.../jvmci/jdk1.8.0_45/product/bin/java -jvmci -d64 -Djvmci.compiler=graal -Xbootclasspath/p:/.../truffle/mxbuild/dists/truffle-api.jar -version
```

Note that the `--vmprefix` comes from [JVMCI](http://hg.openjdk.java.net/graal/graal-jvmci-8/). The help information of the option is also misleading:
```
  --vmprefix <prefix>   prefix for running the VM (e.g. "/usr/bin/gdb --args")
  --gdb                 alias for --vmprefix "/usr/bin/gdb --args"
  --lldb                alias for --vmprefix "lldb --"
```

dougxc:
See #23

PR #23, opened 2016-02-02
Fixed issue where spaces in global options defined by mx extensions caused problems.

dougxc:


dougxc:
This addresses issue #20.

zapster:
It seems like this breaks non-mx option handling:
```
$ mx vm -version
  File "/.../mx.py", line 12287, in <module>
    main()
  File "/.../mx.py", line 12160, in main
    _argParser._parse_cmd_line(_opts, firstParse=True)
  File "/.../mx.py", line 6460, in _parse_cmd_line
    assert ' ' in arg, arg
AssertionError: -version
```

PR #28, opened 2016-02-03
Fix: Graceful fallback when binary dependency not available

sanzinger:
Change implements/fixes the fallback, when binary dependency cannot be found (Download error) to plain source dependency.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -5195,7 +5195,7 @@ class Suite:
             Attempts to locate an existing suite in the local context
             Returns the path to the mx.name dir if found else None
             '''
-            if searchMode == 'binary':
+            if _is_binary_mode():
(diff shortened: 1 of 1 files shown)


mx.py
```diff
@@ -5215,41 +5218,54 @@ def _clone_kwargs():
...
+                # end of search
+                if fail:
                     return None
+            return importMxDir
+
+        importMxDir = _try_clone()
+
+        if  _is_binary_mode() and importMxDir is None:
+            log("Binary import suite '{0}' not found, falling back to source dependency".format(suite_import.name))
+            #_change_to_source_mode()
```
dougxc: Do we need that comment to look like a call to a (non-existent) function?

sanzinger: Bummer, This is just leftover and not needed. Will remove it.

PR #30, opened 2016-02-04
In the gate, run pylint and eclipseformat only on the primary suite.

ansalond:
In the gates, pylint and eclipseformat run on the primary suite and on all imported suites. This is not always desirable since the imported suites should be already checked in their own gates. This pull request executes pylint and eclipseformat only on the primary suite when running the gate.


ansalond:
I closed the issue because the change to eclipseformat is not ok.

PR #35, opened 2016-02-05
In the gate, run pylint and eclipseformat only on the primary suite.

ansalond:
In the gates, pylint and eclipseformat run on the primary suite and on all imported suites. This is not always desirable since the imported suites should be already checked in their own gates. This pull request executes pylint and eclipseformat only on the primary suite when running the gate.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -5990,14 +5990,16 @@ def projects_from_names(projectNames):
     else:
         return [project(name) for name in projectNames]
 
-def projects(opt_limit_to_suite=False):
+def projects(opt_limit_to_suite=False, limit_to_primary=False):
     """
-    Get the list of all loaded projects limited by --suite option if opt_limit_to_suite == True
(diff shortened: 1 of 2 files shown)


dougxc:
We should never have introduced `projects_opt_limit_to_suites()` as a shortcut for `projects(opt_limit_to_suite=True)`. As such, I'd prefer not to  introduce `projects_primary_suite` as a shortcut for `projects(limit_to_primary=True)`.

bhossbach:
It's a great improvement in performance that we currently require. The style issue Doug mentioned can be solved by another push.

dougxc:
Fixing the style issue should be trivial right @ansalond? If so, can you please do it now. Otherwise, please open an issue and then I'll integrate this PR.

woess:
In the future, please hold off on merging pull requests when there are pending review comments. Thanks.

PR #36, opened 2016-02-08
Removed 'projects_primary_suite()' and refactoring.

ansalond:
Following up on #35, I removed 'projects_primary_suite()' and refactored the existing code to use 'projects(opt_limit_to_suite=True)' rather than 'projects_opt_limit_to_suites()'.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -5992,28 +5992,21 @@ def projects_from_names(projectNames):
 
 def projects(opt_limit_to_suite=False, limit_to_primary=False):
     """
-    Get the list of all loaded projects limited by --suite option if opt_limit_to_suite == True or by the primary suite if limit_to_primary == True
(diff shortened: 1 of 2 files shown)


mx.py
```diff
@@ -5992,28 +5992,21 @@ def projects_from_names(projectNames):
...
     if opt_limit_to_suite:
-        return _dependencies_opt_limit_to_suites(sortedProjects)
-    elif limit_to_primary:
-        return _dependencies_limited_to_suites(sortedProjects, [_primary_suite.name])
-    else:
-        return sortedProjects
+        sortedProjects = _dependencies_opt_limit_to_suites(sortedProjects)
+    if limit_to_primary:
+        sortedProjects = _dependencies_limited_to_suites(sortedProjects, [_primary_suite.name])
+    return sortedProjects
```
dougxc: Nice improvement.

PR #38, opened 2016-02-10
Make git revision checking more resilient

christianwimmer:
git show can return non-zero exit code when revision does not exist yet. Instead of aborting, we can just say that the revision does not exist yet.


christianwimmer:
@dougxc please review and merge

PR #39, opened 2016-02-10
Add option to deploy only platform dependent binary distributions.

rschatz:
This adds a `mx deploy-binary --platform-dependent` option, to only deploy platform dependent build artifacts. That way, we can deploy built binaries for e.g. sparc or mac without deploying the platform independent jar files more than once.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -4204,12 +4204,14 @@ def deploy_binary(args):
                             Path to settings.mxl file used for Maven
       -n, --dry-run         Dry run that only prints the action a normal run would
                             perform without actually deploying anything
+      --platform-dependent  Limit deployment to platform dependent distributions only
       --only ONLY           Limit deployment to these distributions
     """
     parser = ArgumentParser(prog='mx deploy-binary')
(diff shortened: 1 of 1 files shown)


mx.py
```diff
@@ -4243,7 +4247,8 @@ def _versionGetter(suite):
 
     version = _versionGetter(s)
     log('Deploying {0} distributions for version {1}'.format(s.name, version))
-    _deploy_binary_maven(s, _map_to_maven_dist_name(mxMetaName), _mavenGroupId(s), mxMetaJar, version, repo.name, repo.url, settingsXml=args.settings, dryRun=args.dry_run)
+    if not args.platform_dependent:
+        _deploy_binary_maven(s, _map_to_maven_dist_name(mxMetaName), _mavenGroupId(s), mxMetaJar, version, repo.name, repo.url, settingsXml=args.settings, dryRun=args.dry_run)
```
dougxc: Why do we not want to deploy binaries when `--platform-dependent` is not given?

rschatz: That line is deploying the "jvmci-mx.jar", which is always platform independent.

dougxc: Ok, makes sense.

PR #40, opened 2016-02-10
Allow dedicated comment field in suite definition

juergenchrist:
This field can be used to specify arbitrary things that are not interpreted by
mx.  This is useful to, e.g., depend on non-mx software like, e.g., a database
label.  Since the feature added in this commit is not backward compatible
(using new suite.py files with an older mx version will have mx complain about
the unknown key 'comment') I increased the minor revision number.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -4777,7 +4777,16 @@ class Suite:
         if javacLintOverrides:
             self.javacLintOverrides = javacLintOverrides.split(',')
 
-        unknown = frozenset(d.keys()) - frozenset(supported)
+        unknown = set(d.keys()) - frozenset(supported)
+
+        suiteExtensionAttributePrefix = self.name + ':'
+        suiteSpecific = {n[len(suiteExtensionAttributePrefix):] : d[n] for n in d.iterkeys() if n.startswith(suiteExtensionAttributePrefix) and n != suiteExtensionAttributePrefix}
+        for n, v in suiteSpecific.iteritems():
+            if hasattr(self, n):
+                abort('Cannot override built-in suite attribute "' + n + '"', context=self)
+            setattr(self, n, v)
+            unknown.remove(suiteExtensionAttributePrefix + n)
+
         if unknown:
             abort(modulePath + ' defines unsupported suite attribute: ' + ', '.join(unknown))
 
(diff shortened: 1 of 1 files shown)


mx.py
```diff
@@ -12311,7 +12317,7 @@ def alarm_handler(signum, frame):
         # no need to show the stack trace when the user presses CTRL-C
         abort(1)
 
-version = VersionSpec("5.6.16")
+version = VersionSpec("5.7.16")
```
woess: bad version bump

dougxc:
I think a better approach is to allow any suite specific attributes. They can simply be those distinguished by a prefix that is the suite's name. For example, in `suite.py` for `graal-core` the :

`"graal-core.required_database_version" = "1.0"`

dougxc:
Here's a variation on the general mechanism I proposed above:

```
diff --git a/mx.py b/mx.py
index d0d1261..d3f3915 100755
--- a/mx.py
+++ b/mx.py
@@ -4761,7 +4761,16 @@ class Suite:
         if javacLintOverrides:
             self.javacLintOverrides = javacLintOverrides.split(',')

-        unknown = frozenset(d.keys()) - frozenset(supported)
+        unknown = set(d.keys()) - frozenset(supported)
+
+        suiteExtensionAttributePrefix = self.name + ':'
+        suiteSpecific = {n[len(suiteExtensionAttributePrefix):] : d[n] for n in d.iterkeys() if n.startswith(suiteExtensionAttributePrefix) and n != suiteExtensionAttributePrefix}
+        for n, v in suiteSpecific.iteritems():
+            if hasattr(self, n):
+                abort('Cannot override built-in suite attribute "' + n + '"', context=self)
+            setattr(self, n, v)
+            unknown.remove(suiteExtensionAttributePrefix + n)
+
         if unknown:
             abort(modulePath + ' defines unsupported suite attribute: ' + ', '.join(unknown))

```

The example would now be:

`"graal-core:required_db_version" : "1.0"`

and assuming `s` is the suite object for `graal-core` the extension attribute is accessed as:

`s.required_db_version`

juergenchrist:
That one would work fine for me.  But I still would to bump the version to specify that this is not backward compatible.

dougxc:
Yes, you still need the version bump. Can you please update this PR to include my patch and the version bump but remove the other changes.

PR #42, opened 2016-02-11
feature: support tagging of gate tasks

zapster:
This change allows the use of _tags_ to specify groups of gate tasks. Every task can be equipped with a set of tags.

The user can select one or more tags using the `--tags` argument of the `mx gate` command. For instance `mx gate --tags style,build` executes all tasks which are tagged with `style` or `build` (logical _or_):
```
$ mx -v gate --dry-run --tags style,build
...
Gate task times:
  0:00:00.000024        Pylint [style]
  0:00:00.000024        Clean [build,style]
  0:00:00.000023        Distribution Overlap Check [style]
  0:00:00.000017        Canonicalization Check [style]
  0:00:00.000017        BuildJavaWithEcj [style]
  0:00:00.000018        BuildJavaWithJavac [build,style]
  0:00:00.000017        IDEConfigCheck [style]
  0:00:00.000029        CodeFormatCheck [style]
  0:00:00.000018        Checkstyle [style]
  0:00:00.000017        Checkheaders [style]
  0:00:00.000016        FindBugs [style]
  =======
  0:00:00.000871
```
Another example:
```
$ mx -v gate --dry-run --tags build
...
Gate task times:
  0:00:00.000032        Clean [build,style]
  0:00:00.000012        BuildJavaWithJavac [build,style]
  =======
  0:00:00.000284
```

It also respects the `-x` flags, which turns it into an exclusion filter. So `mx gate --tags style -x` omits all tasks which _only_ contain the `style` tag (logical _and_).
```
$ mx -v gate --dry-run --tags style -x
...
Gate task times:
  0:00:00.000014        Versions
  0:00:00.000012        JDKReleaseInfo
  0:00:00.000012        Clean [build,style]
  0:00:00.000012        BuildJavaWithJavac [build,style]
  =======
  0:00:00.000284
```
Example with two tags:
```
$ mx -v gate --dry-run --tags style,build -x
...
Gate task times:
  0:00:00.000017        Versions
  0:00:00.000013        JDKReleaseInfo
  =======
  0:00:00.000218
```

There are other possible interpretations for selection and inclusion/exclusion but I think this behavior is the most useful trade-off between functionality and complexity for our use case. This is of course open for discussion.

Note that it can not be used in conjunction with the filter options (`--task-filter`, `--start-at`).

This PR also introduces two tags, `style` and `build`, and applies them to the existing Tasks wherever appropriate. We could also add other _predefined_ tags like `Tags.test`. Note that suites are not restricted to the `Tags` class but can use any string.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -12327,7 +12327,7 @@ def main():
         # no need to show the stack trace when the user presses CTRL-C
         abort(1)
 
-version = VersionSpec("5.6.16")
+version = VersionSpec("5.7.0")
 
 currentUmask = None
 
--- a/mx.py
+++ b/mx.py
@@ -6548,6 +6548,7 @@ class ArgParser(ArgumentParser):
             opts.ignored_projects += os.environ.get('IGNORED_PROJECTS', '').split(',')
 
             mx_gate._jacoco = opts.jacoco
+            mx_gate.Task.verbose = opts.verbose
         else:
             parser = ArgParser(parents=[self])
             parser.add_argument('commandAndArgs', nargs=REMAINDER, metavar='command args...')
--- a/mx_gate.py
+++ b/mx_gate.py
@@ -45,10 +45,14 @@ class Task:
     startAtFilter = None
     filtersExclude = False
 
-    def __init__(self, title, tasks=None, disableJacoco=False):
+    tags = None
+    tagsExclude = False
+
+    def __init__(self, title, tasks=None, disableJacoco=False, tags=None):
         self.tasks = tasks
         self.title = title
         self.skipped = False
+        self.tags = tags
         if tasks is not None:
             for t in tasks:
                 if t.title == title:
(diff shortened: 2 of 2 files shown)


zapster:
I'll also update the mx version if we decide to merge the change.

rschatz:
Based on the previous version bumps, and seeing that this is a backwards compatible change, I'd say bugfix version bump.

Other than that, lgtm.

dougxc:
I like it but I think this is closer to a minor version bump than a bugfix (what bug is being fixed?).

rschatz:
One minor comment: It would be nice to have a special `always` tag that's always implicitly included (unless explicitly removed with `-x`). I would assign this to the `Versions` and `JDKReleaseInfo` tasks.

woess:
I like it, except for the `-x` option (which is somewhat disconnected from `--tags` and `--task-filter`). I'd rather have something along the lines of `--tags=incl,~excl` or `--tags-exclude=excl`.

zapster:
@woess can you provide a real usage example? I tried to keep the interface simple and focused on real use cases.

rschatz:
The problem is, something like `build,~style` doesn't really make any sense.

In the current behavior, `--tags build` implicitly excludes everything that's `style` but not `build`. And it doesn't make sense to run `build` steps but not `build,style` steps, because the `style` steps may depend on the `build,style` steps (e.g. findbugs won't run if the javac build wasn't run before).

woess:
@rschatz Ok, I see.
Sorry, I misinterpreted the meaning of `-x`. I still think the option name could be more explicit about what it does. Maybe not a problem with the current set of options (as there can only be one filter), but it could become ambiguous in the future. For now it's probably ok.

zapster:
@woess right, we should probably add a long-version for `-x`

zapster:
Superseded by #45. Closing.

PR #44, opened 2016-02-11
Added information about versioning

juergenchrist:
README.md now includes some guidelines on how to update the version number.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,4 +40,46 @@ the major as it is, increase the minor number and reset the patch level.
 the current version can still be loaded with the new version and vice versa),
 leave the major and minor versions as they are but increase the patch level.
 
+The version update strategy is designed to help users to detect if their `mx`
+version is compatible with a suite.  Thus, changes to the code that do not
+affect users do not require a change in the version number.  See the following
+examples.  In these examples, by *user* we mean command line clients or `mx`
+extensions (for example `mx_graal-core.py`).
+
+- "I found a for-loop in the code that could be expressed using a map
+function. I changed it accordingly."  This change has no influence on users.
+Thus, no version change is required!
+
+- "I added a new `mx` command."  Since this function was not available to
+users before, old scripts will continue to work with the new version.  New
+scripts, however, might not work with old versions.  This is a minor update
+and requires a new minor number and a reset of the patch level.
+
+- "I fixed a bug that caused a wrong result of a publicly available function."
+This is a bugfix that is user visible.  The patch level should be increased
+since users of old versions can expect at least the bug that was just fixed.
+
+- "I fixed some documentation."  This fix has no impact on the usage of `mx`
+and should thus not change the version of `mx`.
+
+- "I fixed a function.  The result now differs from the results before.  A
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,4 +18,26 @@ If no options or command is specified, `mx` prints information on the available
...
+`mx` uses a `major`.`minor`.`patch` versioning scheme.  To figure out if you
+version is sufficient for a given `mx` suite, first compare the major version
+number of your `mx` version against the major number of the required version
+specified in the suite.  If these versions are not equal, you cannot expect
+`mx` to be compatible with this suite.  The minor number has to be greater or
+equal to the specified minor version number.  Compatibility is ensured
+regardless of the patch level.  However, if your patch level is lower than the
+required patch level you might trigger bugs in `mx`.
+
+From a developer point of view this versioning scheme enforces the following
```
dougxc: from a mx developer point of view

README.md
```diff
@@ -18,4 +18,26 @@ If no options or command is specified, `mx` prints information on the available
 
 For an example of `mx` usage, you can read the [Instructions][1] for the Graal project. 
 
+### mx versioning ###
+
+`mx` uses a `major`.`minor`.`patch` versioning scheme.  To figure out if you
```
dougxc: To figure out if **the** version

dougxc:
lgtm modulo fixes for above comments

rschatz:
Does that mean that *any* change to mx bumps the patch version number? I always was under the impression that pure bugfixes that have no impact on the interface don't need a version bump...

README.md
```diff
@@ -18,4 +18,26 @@ If no options or command is specified, `mx` prints information on the available
...
+`mx` uses a `major`.`minor`.`patch` versioning scheme.  To figure out if the
+version is sufficient for a given `mx` suite, first compare the major version
+number of your `mx` version against the major number of the required version
+specified in the suite.  If these versions are not equal, you cannot expect
+`mx` to be compatible with this suite.  The minor number has to be greater or
+equal to the specified minor version number.  Compatibility is ensured
+regardless of the patch level.  However, if your patch level is lower than the
+required patch level you might trigger bugs in `mx`.
+
+From a `mx` developer point of view this versioning scheme enforces the following
```
woess: shouldn't it be **an** mx developer?

dougxc: I agree that "an mx" sounds better but I believe "a mx" is [correct](http://writingexplained.org/a-vs-an-difference)

dougxc:
It seems hard to come up with a clear policy when to bump the patch level. I think the rough distinction of bug fix versus other minor fixes such as typos fixes, documentation update, private variable renaming variable may be useful. We should increase patch number for the former but not necessarily for the latter.

README.md
rschatz: Actually, the link you posted says exactly the opposite:
> An is used before words, abbreviations, acronyms, or letters that begin with a **vowel sound**, regardless of their spelling

They even give an example that's probably as close as you can get before "mx" finally gets accepted into dictionaries ;)
>A Microsoft program.
>-but-
>An MS-DOS program.

PR #47, opened 2016-02-15
Limit number of cpus and memory

juergenchrist:
On big systems, mx dies since it spawns too many JVMs in parallel.  We now add
an option to limit the number of cpus from the command line.

Also the memory used to detect the Xlint options of javac is limited to prevent
VM crashes on parallel builds.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -104,7 +104,7 @@ def cpu_count():
         cpus = runtime.availableProcessors()
     else:
         cpus = multiprocessing.cpu_count()
-    if hasattr(_opts, 'cpu_count'):
+    if _opts.cpu_count:
         return cpus if cpus <= _opts.cpu_count else _opts.cpu_count
     else:
         return cpus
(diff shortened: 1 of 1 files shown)


mx.py
```diff
@@ -6463,6 +6467,7 @@ def __init__(self, parents=None):
         self.add_argument('--mx-tests', action='store_true', help='load mxtests suite (mx debugging)')
         self.add_argument('--jdk', action='store', help='JDK to use for the "java" command', metavar='<tag:compliance>')
         self.add_argument('--version-conflict-resolution', dest='version_conflict_resolution', action='store', help='resolution mechanism used when a suite is imported with different versions', default='suite', choices=['suite', 'none', 'latest', 'ignore'])
+        self.add_argument('-j', action='store', type=int, dest='cpu_count', help='the maximum number of cpus to use during build', metavar='<cpus>')
```
dougxc: This option should have a long form as well, something like `--cpus`.

mx.py
```diff
@@ -101,9 +101,13 @@ def cpu_count():
     if is_jython():
         from java.lang import Runtime
         runtime = Runtime.getRuntime()
-        return runtime.availableProcessors()
+        cpus = runtime.availableProcessors()
     else:
-        return multiprocessing.cpu_count()
+        cpus = multiprocessing.cpu_count()
+    if hasattr(_opts, 'cpu_count'):
```
dougxc: Why are you using `hasattr` instead of `_opts.cpu_count`? Is this method called before option parsing has happened?

PR #48, opened 2016-02-15
remove unnecessary quoting for args to "git ls-files" invocation

dougxc:


dougxc:
@juergenchrist please check that the version bump in this change complies with the clarified policy

PR #49, opened 2016-02-16
Add support for library var substitution in urls

olpaw:
This allows us to use the following forms:
- "urls" : ["https://hostname.organization.org/files/{path}"]
- "sourceUrls" : ["https://hostname.organization.org/files/{sourcePath}"]

in library definitions (thus saving redundancy in suite.py files)

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -2222,6 +2222,9 @@ class BaseLibrary(Dependency):
         if licenseId:
             self.theLicense = get_license(licenseId, context=self)
 
+    """
+    Return string where references to instance variables from given text string are replaced.
+    """
     def substVars(self, text):
         return text.format(**vars(self))
 
--- a/mx.py
+++ b/mx.py
@@ -2222,6 +2222,9 @@ class BaseLibrary(Dependency):
         if licenseId:
             self.theLicense = get_license(licenseId, context=self)
 
+    def substVars(self, text):
+        return text.format(**vars(self))
+
 """
 A library that is just a resource and therefore not a ClasspathDependency
 """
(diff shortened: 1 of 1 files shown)


mx.py
```diff
@@ -2222,6 +2222,9 @@ def resolveDeps(self):
         if licenseId:
             self.theLicense = get_license(licenseId, context=self)
 
+    def substVars(self, text):
+        return text.format(**vars(self))
```
dougxc: This means the substitute values may not be the values of the suite.py attributes named in `text`. Couldn't this cause confusion?

olpaw: If `text` contains references to instance variables defined in the current library instance they will get resolved (because `kwargs` expansion of `vars(self)` provides the instance variables as proper arguments for the standard format function).

For example if `text` is: "http://foo.bar.com/slavefiles/{sha1}/{path}" and the instance has the following instance variables: `self.path = "myfile.jar"` and `self.sha1 = ''a1ca304d9b6842bddbc160a45150590b82b762e7` then calling `self.substVars("http://foo.bar.com/slavefiles/{sha1}/{path}")` will return `"http://foo.bar.com/slavefiles/a1ca304d9b6842bddbc160a45150590b82b762e7/myfile.jar"`

dougxc: Yes, I understand. It's just that you need to understand you are referring to instance variables which a) may not reflect any attributes in suite.py or b) may have values different from those in suite.py (e.g. a "path" attribute may have a relative path value and a `path` instance variable may be the absolute path derived from that relative path).

This should be added as documentation to `substVars`.

olpaw: Would:
```
+    """
+    Replaces references to instance variables in given text string"""
+    """
+    def substVars(self, text):
+        return text.format(**vars(self))
```
be sufficient?

dougxc: Ok. At some later point when we document the suite.py format, we'll convey this information as well.

olpaw:
@dougxc, ok now?

PR #50, opened 2016-02-18
gate: add separate tag for style checks that require a build

rschatz:
This makes it possible to parallelize gate checks better.

- The `style` tag will not run any build steps, and it will do all style checks that are possible without building.
- The `build` tag will do the minimal work to get a running binary (i.e. clean, javac).
- The `fullbuild` tag does an ecj build, javac build, ideinit and findbugs.

### DIFF
--- a/mx_gate.py
+++ b/mx_gate.py
@@ -35,9 +35,10 @@ import mx
 Predefined Task tags.
 """
 class Tags:
-    always = 'always'
-    style = 'style'
-    build = 'build'
+    always = 'always'       # special tag that is always implicitly selected
+    style = 'style'         # code style checks (without build)
+    build = 'build'         # build
+    fullbuild = 'fullbuild' # full build (including warnings, findbugs and ide init)
 
 """
 Context manager for a single gate task that can prevent the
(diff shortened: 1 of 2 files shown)


mx_gate.py
```diff
@@ -253,7 +254,7 @@ def gate(args):
                 if mx.command_function('pylint')(['--primary']) != 0:
                     _warn_or_abort('Pylint not configured correctly. Cannot execute Pylint task.', args.strict_mode)
 
-        gate_clean(cleanArgs, tasks, tags=[Tags.build, Tags.style])
+        gate_clean(cleanArgs, tasks, tags=[Tags.build, Tags.fullbuild])
```
dougxc: `fullbuild` should really imply `build` but that's probably adding unnecessary complexity to the tags system.

rschatz: I thought about that, too, but I think for now it's simpler to just do this "manually", i.e. every `build` step is also tagged `fullbuild`.

This feature would be more useful in e.g. the Graal gate, where we would like the `test` tag to implicitly select `build`. Anyway, if we really want that, we can do it in a separate change.

Issue #58, opened 2016-02-26
add support for running unit tests in binary dependencies


dougxc:
Recently the graal-core gate switch to using Truffle as a binary dependency. This had the unfortunately side effect of stopping the Truffle unit tests from being run on Graal in the gate. To fix this we need:

- Truffle to export its unit tests as a distribution ([like Graal does](https://github.com/graalvm/graal-core/blob/master/mx.graal-core/suite.py#L1242)).
- `mx unittest` command to look in distribution jars for tests

PR #59, opened 2016-02-28
add support for running unit tests from binary suite dependencies

dougxc:
This addresses #58.

### DIFF
--- a/mx_unittest.py
+++ b/mx_unittest.py
@@ -36,13 +36,39 @@ from os.path import exists, join
 
 def _find_classes_with_annotations(p, pkgRoot, annotations, includeInnerClasses=False):
     """
-    Scan the sources of project 'p' for Java source files containing a line starting with 'annotation'
-    (ignoring preceding whitespace) and return the fully qualified class name for each Java
-    source file matched in a list.
+    Scan the sources of project 'p' for Java source files containing a line starting with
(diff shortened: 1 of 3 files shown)


dougxc:
@chumer can you please try this once you've created the distribution(s) in Truffle packaging up the Truffle tests that should be tested on Graal.

mx_unittest.py
```diff
@@ -36,14 +36,40 @@
...
+    """
+    Scan distributions from binary suite dependencies for classes contain at least one method
+    with an annotation from 'annotations' and return a dictionary from fully qualified class
+    names to the distribution containing the class.
+    """
+    binarySuiteDists = [d for d in mx.dependencies(opt_limit_to_suite=True) if d.isJARDistribution() and
+                        isinstance(d.suite, mx.BinarySuite) and (not suite or suite == d.suite)]
+    if len(binarySuiteDists) != 0:
+        # Ensure Java support class is built
+        mx.build(['--only', 'com.oracle.mxtool.junit'])
```
gilles-duboscq: I think we should use `--dependencies` instead of `--only` to avoid surprises regarding the dependencies of `com.oracle.mxtool.junit`.
For example, with `--only` the build could fail because some classes in `com.oracle.mxtool.junit` depend on the `JUNIT` library which is downloaded when using `--dependencies` but not necessarily when using `--only`.

PR #60, opened 2016-02-29
Compute the min heap size required to a Java program.

ansalond:
I added a command that computes the minimum heap size required to run a Java program within a certain overhead factor.

This required some changes to `run_java` (run with a timeout) and to `_waitWithTimeout` (the timeout of a subprocess should not be fatal).

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -7055,14 +7055,15 @@ def run_java(args, nonZeroIsFatal=True, out=None, err=None, cwd=None, timeout=No
         jdk = get_jdk()
     return jdk.run_java(args, nonZeroIsFatal=nonZeroIsFatal, out=out, err=err, cwd=cwd, timeout=timeout, env=env, addDefaultArgs=addDefaultArgs)
 
-def run_java_min_heap(args, benchName='# MinHeap:', overheadFactor=1.5, referenceHeap=4096, out=None, err=None, cwd=None, timeout=None, env=None, addDefaultArgs=True, jdk=None):
+def run_java_min_heap(args, benchName='# MinHeap:', overheadFactor=1.5, minHeap=0, maxHeap=2048, repetitions=1, out=None, err=None, cwd=None, timeout=None, env=None, addDefaultArgs=True, jdk=None):
     """computes the minimum heap size required to run a Java program within a certain overhead factor"""
+    assert minHeap <= maxHeap
 
     def run_with_heap(heap, args, timeout=timeout, suppressStderr=True, nonZeroIsFatal=False):
-        log('Trying with ' + str(heap) + 'MB of heap...')
+        log('Trying with %sMB of heap...' % heap)
         with open(os.devnull, 'w') as fnull:
             vmArgs, pArgs = extract_VM_args(args=args, useDoubleDash=False, allowClasspath=True, defaultAllVMArgs=True)
-            exitCode = run_java(vmArgs + ['-Xmx' + str(heap) + 'M'] + pArgs, nonZeroIsFatal=nonZeroIsFatal, out=out, err=fnull if suppressStderr else err, cwd=cwd, timeout=timeout, env=env, addDefaultArgs=addDefaultArgs)
+            exitCode = run_java(vmArgs + ['-Xmx%dM' % heap] + pArgs, nonZeroIsFatal=nonZeroIsFatal, out=out, err=fnull if suppressStderr else err, cwd=cwd, timeout=timeout, env=env, addDefaultArgs=addDefaultArgs)
             if exitCode:
                 log('failed')
             else:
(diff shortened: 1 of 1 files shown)


mx.py
```diff
@@ -7089,7 +7126,14 @@ def _returncode(status):
             return _returncode(status)
         remaining = end - time.time()
         if remaining <= 0:
-            abort('Process timed out after {0} seconds: {1}'.format(timeout, ' '.join(args)))
+            msg = 'Process timed out after {0} seconds: {1}'.format(timeout, ' '.join(args))
+            if nonZeroIsFatal:
+                abort(msg)
+            else:
+                log(msg)
+                for p, _ in _currentSubprocesses:
```
gilles-duboscq: Why kill all the sub-processes? why not just `process`?

mx.py
```diff
@@ -7053,6 +7053,43 @@ def run_java(args, nonZeroIsFatal=True, out=None, err=None, cwd=None, timeout=No
         jdk = get_jdk()
     return jdk.run_java(args, nonZeroIsFatal=nonZeroIsFatal, out=out, err=err, cwd=cwd, timeout=timeout, env=env, addDefaultArgs=addDefaultArgs)
 
+def run_java_min_heap(args, benchName='# MinHeap:', overheadFactor=1.5, referenceHeap=4096, out=None, err=None, cwd=None, timeout=None, env=None, addDefaultArgs=True, jdk=None):
+    """computes the minimum heap size required to run a Java program within a certain overhead factor"""
+
+    def run_with_heap(heap, args, timeout=timeout, suppressStderr=True, nonZeroIsFatal=False):
+        log('Trying with ' + str(heap) + 'MB of heap...')
+        with open(os.devnull, 'w') as fnull:
+            exitCode = run_java(['-Xmx' + str(heap) + 'M'] + args, nonZeroIsFatal=nonZeroIsFatal, out=out, err=fnull if suppressStderr else err, cwd=cwd, timeout=timeout, env=env, addDefaultArgs=addDefaultArgs)
```
gilles-duboscq: I think the last `-Xmx` arg is taken into account, maybe you could try to insert as the last JVM arg? I think only the `-cp` and `--classpath` JVM arg consume the next argument.

mx.py
```diff
@@ -7089,7 +7126,14 @@ def _returncode(status):
...
         if remaining <= 0:
-            abort('Process timed out after {0} seconds: {1}'.format(timeout, ' '.join(args)))
+            msg = 'Process timed out after {0} seconds: {1}'.format(timeout, ' '.join(args))
+            if nonZeroIsFatal:
+                abort(msg)
+            else:
+                log(msg)
+                for p, _ in _currentSubprocesses:
+                    _kill_process_group(p.pid, signal.SIGKILL)
+                return 1
```
gilles-duboscq: Maybe you can return something which is not a valid return code so it's possible to distinguish failed runs from timeouts? (e.g., `ERROR_TIMEOUT = 0x700000000 # not 32 bits`)

mx.py
```diff
@@ -7053,6 +7053,43 @@ def run_java(args, nonZeroIsFatal=True, out=None, err=None, cwd=None, timeout=No
...
+            return 1
+        referenceTime = time.time() - t
+        maxTime = referenceTime * overheadFactor
+        log('Reference time = ' + str(referenceTime))
+        log('Maximum time = ' + str(maxTime))
+    else:
+        maxTime = None
+
+    heap = 10
+    while run_with_heap(heap, args, timeout=maxTime):
```
gilles-duboscq: if you do the `ERROR_TIMEOUT` stuff below then maybe you should check if we get `ERROR_TIMEOUT` here and abort if you get something else which is not `0`.

mx.py
ansalond: `abort` would kill all sub-processes. I tried to keep the semantics as similar as possible, so that the only difference is that the current process survives. However, I see your point and I am going to change this in my next push.

mx.py
ansalond: Good point. I will call `extract_VM_args` and append `-Xmx` to that list.

mx.py
ansalond: This loop increases the heap size to be used in the next attempt every time that the process does not return `0`. I don't think it should make any difference if the process timed out or it failed differently (e.g., `OutOfMemoryError` or `Excessive GC Time`).

mx.py
ansalond: You are right. I don't think it matters in my use case but it might be important to others.

mx.py
gilles-duboscq: Right

mx.py
gilles-duboscq: Do you want to do that one in a separate pull request?

mx.py
Peter-B-Kessler: Do you want to start with some reasonable heap size and then binary search your way down to near the minimum, rather than (as I read it) starting with a heap size that's too small (10MB?) and growing the heap by a fixed amount (10MB then 1MB?) after each failure?  How many iterations does it typically take you to find the minimum?  How does that number of iterations compare with log2(referenceHeap)?  How long does each iteration take for benchmarks you care about?

ansalond: I took this part from a script that we use in a different context and I simply translated it to python. In my use cases the heap sizes that I am measuring range from 5MB to 20MB, and the whole measuring process takes at most 4 mins, so the current algorithm is not so bad. However, I see your point and I will improve the algorithm.

PR #27, opened 2016-02-03
use checkstyle version 6.15

mrigger:
The changes address issue #25. Suites that rely on older mx version will use Checkstyle 6.0, while newer ones will use 6.15.

PR #62, opened 2016-03-02
Enabling check that every element in the API has its @since tag

jtulach:
Truffle repository requires that every element in the API has its own **@since** tag since https://github.com/graalvm/truffle/pull/86. To make sure we don't regress in this respect, we need to turn on some automatic check.

This proposal turns on printing of warnings for `mx javadoc --unified` when a **@since** tag is missing. The warnings are then interpreted by the Truffle mx customizations as errors and fail the build. Exactly what we need.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -10694,6 +10694,7 @@ def javadoc(args, parser=None, docDir='javadoc', includeDeps=True, stdDoclet=Tru
              '-doclet', 'org.apidesign.javadoc.codesnippet.Doclet',
              '-docletpath', snippetslib,
              '-snippetpath', snippets,
+             '-verifysincepresent',
              '-sourcepath', sp] +
              ([] if jdk.javaCompliance < JavaCompliance('1.8') else ['-Xdoclint:none']) +
              (['-overview', overviewFile] if exists(overviewFile) else []) +
--- a/mx.py
+++ b/mx.py
@@ -10522,9 +10522,11 @@ def javadoc(args, parser=None, docDir='javadoc', includeDeps=True, stdDoclet=Tru
             if not added:
                 logv('[{0} - skipping {1}]'.format(reason, p.name))
     snippets = []
+    verifySincePresent = [];
     for p in projects_opt_limit_to_suites():
         if p.isJavaProject():
             snippets += p.source_dirs()
+            verifySincePresent = p.suite.getMxCompatibility().verifySincePresent()
     snippets = os.pathsep.join(snippets)
     snippetslib = library('CODESNIPPET-DOCLET').get_path(resolve=True)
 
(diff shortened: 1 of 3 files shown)


jtulach:
Who could do the review? @gilles-duboscq?

mx.py
```diff
@@ -10694,6 +10694,7 @@ def __call__(self, msg):
              '-doclet', 'org.apidesign.javadoc.codesnippet.Doclet',
              '-docletpath', snippetslib,
              '-snippetpath', snippets,
+             '-verifysincepresent',
```
zapster: I guess you should also add the same change to [line 10600](https://github.com/graalvm/mx/pull/62/files#diff-1850d7da100ab46a504d0dd5236dc495R10600), right?

jtulach: The `mx javadoc --unified` is used to produce documentation for API - probably only by Truffle. As far as I understand `mx javadoc` is used for documentation for the code - to help people orient around the project source. Probably the `-verifysincepresent` check is only useful for the documentation of an API and would be annoying otherwise.

jtulach:
Doug suggested to try Roland too...

rschatz:
This is an API-breaking change, right? It's going to report problems when it didn't before.

So this needs a version bump, and `mx_compat` should be used (i.e. the `-verifysincepresent` should only be passed to `javadoc` if the suite has `mxversion` pointing to this version or higher). Otherwise we're going to break old truffle builds. See #27 for an example of how to do this.

jtulach:
If you give me go, I squash the commits into one and integrate.

rschatz:
Please also bump the version to `5.9.0`:
```diff
diff --git a/mx.py b/mx.py
index 1b7175e..db90a92 100755
--- a/mx.py
+++ b/mx.py
@@ -12429,7 +12429,7 @@ def main():
         # no need to show the stack trace when the user presses CTRL-C
         abort(1)
 
-version = VersionSpec("5.8.1")
+version = VersionSpec("5.9.0")
 
 currentUmask = None
 
diff --git a/mx_compat.py b/mx_compat.py
index e81e67d..bfad342 100644
--- a/mx_compat.py
+++ b/mx_compat.py
@@ -142,10 +142,10 @@ class MxCompatibility5616(MxCompatibility566):#pylint: disable=too-many-ancestor
     def checkstyleLibraryName(self):
         return 'CHECKSTYLE_6.15'
 
-class MxCompatibility62(MxCompatibility5616):#pylint: disable=too-many-ancestors
+class MxCompatibility59(MxCompatibility5616):#pylint: disable=too-many-ancestors
     @staticmethod
     def version():
-        return mx.VersionSpec("5.8.1")
+        return mx.VersionSpec("5.9.0")
 
     def verifySincePresent(self):
         return ['-verifysincepresent']
```

Aside from that, lgtm.

PR #63, opened 2016-03-02
add support for suite import URL rewriting

dougxc:
This adds support for a suite to be able to override the source URLs of its imported suites. In particular, a new suite level `urlrewrites` attribute is added. This allows regular expression URL rewriting as in this example:

```
  "urlrewrites" : [
    { 
      "https://git.acme.com/(.*).git" : {
        "replacement" : r”https://my.company.com/foo-git-cache/\1.git",
      }
    },
    { 
      "https://hg.acme.com/(.*)" : {
        "replacement" : r”https://my.company.com/foo-hg-cache/\1",
      }
    }
  ],
```

The rules are applied in definition order. Only rewrite rules from the primary suite are used meaning a suite may have to replicate the rewrite rules of its suite dependencies.

Note that this only adds support for (source and binary) suite import URLs. A possible follow up may be to add support for rewriting other URLs such as library URLs. In this case, an extra `scope` attribute could be added for each rewrite rule specifying which URLs should be rewritten. If library URL rewrite support is added, then support for sha1 rewriting may also be required since rewritten library URLs may point to binaries that were built differently from the same source.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -154,6 +154,7 @@ _vc_systems = []
 _mvn = None
 _binary_suites = None # source suites only if None, [] means all binary, otherwise specific list
 _suites_ignore_versions = None # as per _binary_suites, ignore versions on clone
+_urlrewrites = [] # list of URLRewrite objects
 
 # List of functions to run when the primary suite is initialized
 _primary_suite_deferrables = []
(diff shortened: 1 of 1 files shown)


gilles-duboscq:
:+1:

PR #64, opened 2016-03-04
Individual IDE init routines should do fsckprojects

tkrodriguez:
ideinit does an fsckprojects at the end but the IDE specific routines do not and they should.


rschatz:
Even though it's a bugfix, this has a user-visible effect, so this needs a patch level version bump.
Other than that, lgtm.

tkrodriguez:
Added version bump

Issue #57, opened 2016-02-26
Support embedded snippets in Truffle


jtulach:
There is a discussion in https://github.com/graalvm/truffle/issues/19 to improve co-location of code snippets to the API that they describe. There are few needed changes in **mx** to make to current proposal functional:
- upgrade to [new version of the doclet](https://github.com/jtulach/codesnippet4javadoc/commit/515fdd141c8caed9d86afce859afb15a81054f7f) (once released)
- add additional parameters to javadoc to specify pattern for files that contain the snippets
-- in *Truffle* they contain letters *Snippets*
- filter out these files from final JARs

Here is the diff that achieves all of that. Can you please tell me how to polish it to be ready for integration into mx repository? Thanks.

```diff
diff --git a/mx.mx/suite.py b/mx.mx/suite.py                                                                                                                                                            
index 679dc5b..b9afd6e 100644                                                                                                                                                                           
--- a/mx.mx/suite.py
+++ b/mx.mx/suite.py
diff --git a/mx.mx/suite.py b/mx.mx/suite.py
index 679dc5b..b9afd6e 100644
--- a/mx.mx/suite.py
+++ b/mx.mx/suite.py
@@ -31,9 +31,9 @@ suite = {
 
     "CODESNIPPET-DOCLET" : {
       "urls" : [
-        "http://repo1.maven.org/maven2/org/apidesign/javadoc/codesnippet-doclet/0.7/codesnippet-doclet-0.7.jar",
+        "file:///home/devel/.m2/repository/org/apidesign/javadoc/codesnippet-doclet/1.0-SNAPSHOT/codesnippet-doclet-1.0-SNAPSHOT.jar",
       ],
-      "sha1" : "28aaf1920d83e1e7d994c50cd23576ef93195434",
+      "sha1" : "fbf655c3f3abdd3d10970f0c5ecf062b6c3c7dc8",
     },
 
     "JUNIT" : {
diff --git a/mx.py b/mx.py
index 9ae0f6b..99fc75c 100755
--- a/mx.py
+++ b/mx.py
@@ -964,6 +964,8 @@ class JARDistribution(Distribution, ClasspathDependency):
                                         services.setdefault(service, []).extend([provider.strip() for provider in fp.readlines()])
                             else:
                                 for f in files:
+                                    if ('Snippets' in f):
+                                        continue
                                     arcname = join(relpath, f).replace(os.sep, '/')
                                     with open(join(root, f), 'rb') as fp:
[...]

PR #65, opened 2016-03-04
Include codesnippets into the Javadoc and exclude them from distribut…

jtulach:
…ion JARs if specified as `'snippetsPattern' : '.*SampleSnippets.*'` in suite.py

Pull request to implement what has been discussed in https://github.com/graalvm/mx/issues/57

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -879,6 +879,9 @@ class JARDistribution(Distribution, ClasspathDependency):
     def make_archive(self):
         # are sources combined into main archive?
         unified = self.path == self.sourcesPath
+        snippetsPattern = None
+        if hasattr(self.suite, 'snippetsPattern'):
+            snippetsPattern = re.compile(self.suite.snippetsPattern)
 
         with Archiver(self.path) as arc:
             with Archiver(None if unified else self.sourcesPath) as srcArcRaw:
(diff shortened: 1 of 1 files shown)


mx.py
```diff
@@ -4772,9 +4777,8 @@ def expand(value, context):
             'developer',
             'url',
             'licenses',
-            'licences',
             'defaultLicense',
-            'defaultLicence',
```
rschatz: Why did you remove these two keys? What effect does removing them have?
(I'm not familiar with that part of the code... @gilles-duboscq, do you know what this does?)

gilles-duboscq: Initially we were using the UK spelling. In an mx version we switched to US spelling. There is `mx_compat` support for that so that we can still load older suites.
I don't think we should remove these keys.

If there is a problem such as mx silently ignoring the old version of the keys, then we need a different solution were we first read the `mxversion` and then add the keys with the current spelling to the list of supported keys using `mx_compat`. But that would anyway be work for an other pull request.

jtulach: I am sorry guys. I thought the keys are duplicated. I haven't noticed the difference in "s" vs. "c". I update my diff.

jtulach:
Now the diff contains only changes related to code snippets. Sorry for the noise and thanks for catching the error.

rschatz:
Please also add a version bump, so users will get a "please update mx" message instead of a cryptic "unsupported suite attribute: snippetsPattern".

jtulach:
All comments addressed. Merging.

PR #67, opened 2016-03-07
Add support for transitive deploy-binary

olpaw:
This change-set provides improvements to the existing `mx deploy-binary` command.
- `deploy-binary` now uses the same positional arguments as `maven-deploy`.
```
  repository-id         Repository ID used for binary deploy
  repository-url        Repository URL used for binary deploy, if no url is
                        given, the repository-id is looked up in suite.py
```
- `deploy-binary` accepts `--all-suites` to perform deployment for all dependent suites (not just primary).
- `deploy-binary` accepts `--skip-existing` to prevent deploying if the suite (of a given revision) was already deployed before.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -2094,6 +2094,17 @@ def _get_path_in_cache(name, sha1, urls, ext=None):
     assert os.pathsep not in name, name + ' cannot contain ' + os.pathsep
     return join(cacheDir, name + '_' + sha1 + ext)
 
+def download_file_exists(urls):
+    '''
+    Returns true if the given urls really exist on the server.
+    '''
+    log('download_file_exists: ' + ', '.join(urls))
+    _, binDir = _compile_mx_class('URLConnectionDownload')
+    command = [get_jdk(tag=DEFAULT_JDK_TAG).java, '-cp', _cygpathU2W(binDir), 'URLConnectionDownload']
+    command += ['--no-progress', '--skip-download']
+    command += urls
+    return run(command, nonZeroIsFatal=False) == 0
+
 def download_file_with_sha1(name, path, urls, sha1, sha1path, resolve, mustExist, sources=False, canSymlink=True):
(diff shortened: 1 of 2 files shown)


ldaynes:
Does that means that we can start trying binary deployment ?

Laurent

olpaw:
@ldaynes, it's a necessary, but not sufficient change. That said, it brings us very close to it.

mx.py
```diff
@@ -4223,51 +4239,89 @@ def deploy_binary(args):
...
 
-    mxMetaName = _mx_binary_distribution_root(s.name)
-    s.create_mx_binary_distribution_jar()
-    mxMetaJar = s.mx_binary_distribution_jar_path()
+    mxMetaName = _mx_binary_distribution_root(suite.name)
+    suite.create_mx_binary_distribution_jar()
+    mxMetaJar = suite.mx_binary_distribution_jar_path()
     assert exists(mxMetaJar)
+    if args.all_suites:
+        dists = [d for d in dists if d.exists()]
```
gilles-duboscq: Why do we ignore missing distributions in `--all-suites` mode?

olpaw: For example, some project transitively depends on JVM_PRODUCT_SERVER but does not require JVM_PRODUCT_JVMCI. Thus there is no need to deploy JVM_PRODUCT_JVMCI.

gilles-duboscq: Well, but then shouldn't we check that instead of checking if the distribution exists?
Otherwise this disables the safety guard that all the necessary distributions will be uploaded.

Also, this means the actual semantics of `--all-suites` is "this suites and the distributions it depends on in other suites".

olpaw: I agree. The help text of `--all-suites` should be changed to `Deploy suite and the distributions it depends on in other suites`. I will create another pull request for that. Thanks for the feedback!

PR #70, opened 2016-03-08
Fixes for deploy-binary --skip-existing and --all-suites

olpaw:
Based on @gilles-duboscq  suggestions I rephrased the help text of the --all-suites option. The other changes make the --skip-existing option operate on distribution- (instead of suite-) granularity.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -4314,13 +4314,13 @@ def _deploy_binary(args, suite):
         repo = repository(args.repository_id)
 
     version = _versionGetter(suite)
-    log('Deploying suite {0} distributions for version {1}'.format(suite.name, version))
+    log('Deploying suite {0} version {1}'.format(suite.name, version))
     if args.skip_existing:
         non_existing_dists = []
         for dist in dists:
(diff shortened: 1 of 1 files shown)


mx.py
```diff
@@ -4313,11 +4314,17 @@ def _versionGetter(suite):
...
     version = _versionGetter(suite)
-    log('Deploying {0} distributions for version {1}'.format(suite.name, version))
+    log('Deploying suite {0} version {1}'.format(suite.name, version))
     if args.skip_existing:
-        metadata_url = '{0}/{1}/{2}/{3}/maven-metadata.xml'.format(repo.url, _mavenGroupId(suite).replace('.', '/'), _map_to_maven_dist_name(mxMetaName), version)
-        if download_file_exists([metadata_url]):
-            log('Skip deployment for already deployed suite {0} for version {1}'.format(suite.name, version))
+        non_existing_dists = []
+        for dist in dists:
+            metadata_url = '{0}/{1}/{2}/{3}/maven-metadata.xml'.format(repo.url, _mavenGroupId(suite).replace('.', '/'), dist.maven_artifact_id(), version)
```
gilles-duboscq: regarding the group ID, you should probably use `dist.maven_group_id()`

olpaw: Ok. Updated to use `dist.maven_group_id()` instead.

PR #72, opened 2016-03-08
Add javac daemon mode to speed up build

tkrodriguez:
@gilles-duboscq this is the basic version of the javac daemon stuff.  I'm not entirely happy with how it integrates with mx so any feedback is appreciated.  I'm also not sure I'm missing anything about special handling of javac versions.  mx allocates a server port and hands it into the javac daemon so that we don't have port number collisions.  That's some ugly code.  It tries to cleanly shutdown the daemon process in the build but I suspect I mean want to set daemon on the process so that it will be forcibly shutdown if the main process exits.  I've got a test build running with the daemon at https://travis-ci.org/tkrodriguez/graal-core/builds/114622495.  I've only done light local testing so far.

### DIFF
--- a/java/com.oracle.mxtool.compilerserver/src/com/oracle/mxtool/compilerserver/JavacDaemon.java
+++ b/java/com.oracle.mxtool.compilerserver/src/com/oracle/mxtool/compilerserver/JavacDaemon.java
@@ -1,0 +1,123 @@
+/*
+ * Copyright (c) 2016, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package com.oracle.mxtool.compilerserver;
+
+import java.io.BufferedReader;
+import java.io.InputStreamReader;
+import java.io.OutputStreamWriter;
+import java.lang.reflect.Method;
+import java.net.InetSocketAddress;
+import java.net.ServerSocket;
+import java.net.Socket;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+public class JavacDaemon implements Runnable {
+
+    private static void logf(String commandLine, Object... args) {
+        if (verbose) {
+            System.err.printf(commandLine, args);
+        }
+    }
+
+    private static boolean verbose = false;
+    private static boolean running;
+    private static ThreadPoolExecutor threadPool;
+
+    private Socket connectionSocket;
+    private Object compiler;
+    private Method compile;
+
+    public static void main(String[] args) throws Exception {
+        if (args.length != 1) {
+            System.err.println("Usage: port");
+            System.exit(1);
+        }
+
+        // create socket
+        int port = Integer.parseInt(args[0]);
+        ServerSocket serverSocket = new ServerSocket();
+        serverSocket.setReuseAddress(true);
+        serverSocket.bind(new InetSocketAddress(port));
+        logf("Started server on port %d\n", port);
+
+        final Class<?> javacMainClass = Class.forName("com.sun.tools.javac.Main");
+        final Method compileMethod = javacMainClass.getMethod("compile", new Class<?>[]{(new String[]{}).getClass()});
+
+        int threadCount = Runtime.getRuntime().availableProcessors();
+        threadPool = new ThreadPoolExecutor(threadCount, threadCount, 0L, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>(), new ThreadFactory() {
+            public Thread newThread(Runnable runnable) {
+                return new Thread(runnable);
+            }
+        });
+
+        running = true;
+        while (running) {
+            Socket connectionSocket = serverSocket.accept();
+            threadPool.submit(new JavacDaemon(connectionSocket, javacMainClass.newInstance(), compileMethod));
+        }
+        serverSocket.close();
+    }
+
+    public JavacDaemon(Socket connectionSocket, Object compiler, Method compile) {
+        this.connectionSocket = connectionSocket;
+        this.compiler = compiler;
+        this.compile = compile;
+    }
+
+    public void run() {
+        try {
+            BufferedReader input = new BufferedReader(new InputStreamReader(connectionSocket.getInputStream(), "UTF-8"));
+            OutputStreamWriter output = new OutputStreamWriter(connectionSocket.getOutputStream(), "UTF-8");
+
+            try {
+                String commandLine = input.readLine();
+                if (commandLine.length() == 0) {
+                    logf("Shutting down\n");
+                    running = false;
+                    while (threadPool.getActiveCount() > 1) {
+                        threadPool.awaitTermination(50, TimeUnit.MILLISECONDS);
+                    }
+                    System.exit(0);
+                } else {
+                    logf("Compiling %s\n", commandLine);
+
+                    Integer result = (Integer) compile.invoke(compiler, new Object[]{commandLine.split(" ")});
+                    logf("Result = %d\n", result);
+
+                    output.write(result + "\n");
+                }
+            } finally {
+                // close IO streams, then socket
+                logf("Closing connection with client\n");
+                output.close();
+                input.close();
+                connectionSocket.close();
+            }
+        } catch (Exception ioe) {
+            ioe.printStackTrace();
+        }
+    }
+}
--- a/mx.py
+++ b/mx.py
@@ -650,6 +650,15 @@ class BuildTask(object):
     def cleanForbidden(self):
         return False
 
+    def prepare(self, daemons):
+        """
+        Perform any task initialization that must be done in the main process.
+        This will be called just befor the task is launched.
+        The 'daemons' argument is a dictionary for storing any persistent state
+        that might be shared between tasks.
+        """
+        pass
+
     """
     Build the artifacts.
     """
@@ -1717,15 +1726,21 @@ class JavaBuildTask(ProjectBuildTask):
         if self.args.jdt and not self.args.force_javac:
             return ECJCompiler(self.args.jdt, self.args.extra_javac_args)
         else:
-            return JavacCompiler(self.args.alt_javac, self.args.extra_javac_args)
+            if self.args.no_daemon or self.args.alt_javac:
+                return JavacCompiler(self.args.alt_javac, self.args.extra_javac_args)
+            else:
+                return JavacDaemonCompiler(self.args.extra_javac_args)
+
+    def prepare(self, daemons):
+        self.compiler = self._getCompiler()
+        self.compiler.prepare(self.jdk, daemons)
 
     def build(self):
-        compiler = self._getCompiler()
         outputDir = self.subject.output_dir()
         ensure_dir_exists(outputDir)
         # Java build
         if self._javaFileList():
-            compiler.build(
+            self.compiler.build(
                 sourceFiles=[_cygpathU2W(f) for f in self._javaFileList()],
                 project=self.subject,
                 jdk=self.jdk,
@@ -1798,6 +1813,13 @@ class JavaCompiler:
     def build(self, sourceFiles, project, jdk, compliance, outputDir, classPath, processorPath, sourceGenDir,
         disableApiRestrictions, warningsAsErrors, showTasks):
         nyi('build', self)
+
+    def prepare(self, jdk, daemons):
+        """
+        Perform any task initialization that must be done in the main process.
+        This will be called just befor the task is launched.
+        """
+        pass
 
 class JavacLikeCompiler(JavaCompiler):
     def __init__(self, extraJavacArgs):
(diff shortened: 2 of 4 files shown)


mx.py
```diff
@@ -1893,11 +1905,84 @@ def buildJavacLike(self, jdk, project, jvmArgs, javacArgs, disableApiRestriction
...
+
+    def run(self, jdk, jvmArgs, javacArgs):
+        # ignore the jvmArgs
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        s.connect(('127.0.0.1', self.daemon.port))
+        s.send(" ".join(javacArgs) + "\n")
+        f = s.makefile()
+        retcode = int(f.readline())
+        s.close()
+        nonZeroIsFatal = True
```
sanzinger: nonZeroIsFatal seems to be unnecessary

mx.py
```diff
@@ -1893,11 +1905,84 @@ def buildJavacLike(self, jdk, project, jvmArgs, javacArgs, disableApiRestriction
...
-        javac = self.altJavac if self.altJavac else jdk.javac
         jvmArgs += jdk.java_args
+        self.run(jdk, jvmArgs, javacArgs)
+
+    def run(self, jdk, jvmArgs, javacArgs):
+        javac = self.altJavac if self.altJavac else jdk.javac
         cmd = [javac] + ['-J' + arg for arg in jvmArgs] + javacArgs
         run(cmd)
 
+class JavacCompilerDaemon(JavacCompiler):
```
sanzinger: Minor: Got confused by the name "JavacCompilerDaemon" - it suggests, that this class represents the javac deamon - but it doesn't. In my opoininon changing it to JavacDaemonCompiler would help, this name suggests it is a compiler which uses the JavacDaemon.

mx.py
```diff
@@ -1794,6 +1803,9 @@ def build(self, sourceFiles, project, jdk, compliance, outputDir, classPath, pro
         disableApiRestrictions, warningsAsErrors, showTasks):
         nyi('build', self)
 
+    def prepare(self, jdk, state):
```
dougxc: What is this for? In particular, when is it called? Please add some documentation.

mx.py
```diff
@@ -649,6 +649,9 @@ def buildForbidden(self):
     def cleanForbidden(self):
         return False
 
+    def prepare(self, daemons):
```
dougxc: Please document this function.

mx.py
```diff
@@ -1893,11 +1905,84 @@ def buildJavacLike(self, jdk, project, jvmArgs, javacArgs, disableApiRestriction
...
+    def runDaemon(self):
+        run([self.jdk.java] + ['-cp', self.jdk.toolsjar + ':' + self.coreCp, self.mainClass, str(self.port)])
+
+    def shutdown(self):
+        try:
+            s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            s.connect(('127.0.0.1', self.port))
+            s.send("\n")
+            s.close()
+        except OSError:
```
dougxc: We probably want to at least `logv` this error.

mx.py
```diff
@@ -1893,11 +1905,84 @@ def buildJavacLike(self, jdk, project, jvmArgs, javacArgs, disableApiRestriction
...
+            build(['--only', 'com.oracle.mxtool.compilerserver', '--no-javac-daemon'])
+        self.coreCp = classpath(['com.oracle.mxtool.compilerserver'])
+        # allocate a new port
+        serversocket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        serversocket.bind((socket.gethostname(), 0))
+        self.port = serversocket.getsockname()[1]
+        serversocket.close()
+
+        self.proc = multiprocessing.Process(target=self.runDaemon, args=())
+        self._finished = False
```
dougxc: Where is `_finished` read?

mx.py
```diff
@@ -1893,11 +1905,84 @@ def buildJavacLike(self, jdk, project, jvmArgs, javacArgs, disableApiRestriction
...
+                    log('[exit code: ' + str(retcode) + ']')
+            abort(retcode)
+
+        return retcode
+
+    def prepare(self, jdk, daemons):
+        name = 'javac-daemon-' + jdk.java
+        self.daemon = daemons.get(name)
+        if not self.daemon:
+            self.daemon = JavacDaemon(jdk)
```
dougxc: Is there any sensible error handling we should do if creation of the daemon fails like fall back to non-daemon compilation?

java/com.oracle.mxtool.compilerserver/src/com/oracle/mxtool/compilerserver/JavacDaemon.java
```diff
@@ -0,0 +1,118 @@
...
+        }
+
+        // create socket
+        int port = Integer.parseInt(args[0]);
+        ServerSocket serverSocket = new ServerSocket();
+        serverSocket.setReuseAddress(true);
+        serverSocket.bind(new InetSocketAddress(port));
+        logf("Started server on port %d\n", port);
+
+        final Class<?> javacMainClass = Class.forName("com.sun.tools.javac.Main");
```
gilles-duboscq: What was the reason for not linking against `tools.jar`? could you document it here?

mx.py
```diff
@@ -1893,11 +1905,84 @@ def buildJavacLike(self, jdk, project, jvmArgs, javacArgs, disableApiRestriction
...
+class Daemon:
+    def shutdown(self):
+        pass
+
+class JavacDaemon(Daemon):
+    def __init__(self, jdk):
+        self.jdk = jdk
+        self.mainClass = 'com.oracle.mxtool.compilerserver.JavacDaemon'
+        if not exists(join(project('com.oracle.mxtool.compilerserver').output_dir(), self.mainClass.replace('.', os.sep) + '.class')):
+            build(['--only', 'com.oracle.mxtool.compilerserver', '--no-javac-daemon'])
```
gilles-duboscq: Can we use `--dependencies` instead of `--only`?
Just in case this project starts to have dependencies.

gilles-duboscq: Also, do you think we could call `build` that unconditionally?
Then we can delegate the responsibility of the "is this built/up-to-date" logic to `build`.

java/com.oracle.mxtool.compilerserver/src/com/oracle/mxtool/compilerserver/JavacDaemon.java
```diff
@@ -0,0 +1,118 @@
...
+        this.compile = compile;
+    }
+
+    public void run() {
+        try {
+            BufferedReader input = new BufferedReader(new InputStreamReader(connectionSocket.getInputStream()));
+            DataOutputStream output = new DataOutputStream(connectionSocket.getOutputStream());
+
+            try {
+                String commandLine = input.readLine();
```
gilles-duboscq: I hope Java and python will agree on the charset used to serialize/deserialize those strings.

mx.py
```diff
@@ -1713,15 +1716,21 @@ def _getCompiler(self):
         if self.args.jdt and not self.args.force_javac:
             return ECJCompiler(self.args.jdt, self.args.extra_javac_args)
         else:
-            return JavacCompiler(self.args.alt_javac, self.args.extra_javac_args)
+            if self.args.no_javac_daemon:
+                return JavacCompiler(self.args.alt_javac, self.args.extra_javac_args)
+            else:
+                return JavacCompilerDaemon(self.args.alt_javac, self.args.extra_javac_args)
```
gilles-duboscq: Maybe we should abort if `alt_javac` is set with a message saying that `--no-javac-daemon` needs to be set when using `--alt-javac`.
Alternatively we can set that up automatically after parsing arguments in `build` by setting `no_javac_daemon` to true if `alt_javac` is true (with a warning maybe)

mx.py
```diff
@@ -1893,11 +1905,84 @@ def buildJavacLike(self, jdk, project, jvmArgs, javacArgs, disableApiRestriction
...
+        serversocket.close()
+
+        self.proc = multiprocessing.Process(target=self.runDaemon, args=())
+        self._finished = False
+        self.proc.daemon = True
+        self.proc.start()
+        self.sub = _addSubprocess(self.proc, [str(self)])
+
+    def runDaemon(self):
+        run([self.jdk.java] + ['-cp', self.jdk.toolsjar + ':' + self.coreCp, self.mainClass, str(self.port)])
```
gilles-duboscq: If you had made `com.oracle.mxtool.compilerserver` depends on `tools.jar` (using `jdkLibraries`) you wouldn't need to do that manually here.

java/com.oracle.mxtool.compilerserver/src/com/oracle/mxtool/compilerserver/JavacDaemon.java
woess: good point. imho, we should force utf-8 on both ends.

mx.py
tkrodriguez: deleted.  That was a copy parse from elsewhere I never revisited.

mx.py
tkrodriguez: Yes that's a better name.

mx.py
tkrodriguez: Maybe.  It's seeming a little fragile at the moment.  mac and linux seem to behave differently so I'm hitting some problems with getting the travis build working.

java/com.oracle.mxtool.compilerserver/src/com/oracle/mxtool/compilerserver/JavacDaemon.java
tkrodriguez: What do you mean by linking here?  I mostly just replicated the way that junit was run with manual construction of the proper command line.  Also I don't think JDK9 has a tools.jar anymore.

mx.py
tkrodriguez: I'll switch to --dependencies and make it unconditional.  I was just following what junit was doing since I didn't know what the right way to handle this was.

java/com.oracle.mxtool.compilerserver/src/com/oracle/mxtool/compilerserver/JavacDaemon.java
tkrodriguez: I'll add that.

mx.py
tkrodriguez: What is alt_javac supposed to be used for?  It would probably make sense to simply force no_javac_daemon is alt_javac is set.

mx.py
tkrodriguez: So you mean reference it in the suite.py?  I think that won't work with JDK9.  javac appears to be part of the core now so we'll have to build the command line differently for that case.

Also I probably should set my compliance to 1.7 since I'm may be using this with a 1.7 JDK.  Right?

java/com.oracle.mxtool.compilerserver/src/com/oracle/mxtool/compilerserver/JavacDaemon.java
gilles-duboscq: I mean not using reflection.
Well there must be something in JDK9 that we can use to compile and run code that uses `com.sun.tools.javac.Main`.

mx.py
gilles-duboscq: 1.7 is probably a good idea yes.
But again, even if it changes in 9, in 8 we can use `jdkLibrary` to reference jars from the JDK. In 9, if it's part of the core library then i guess we don't need anything at all.

mx.py
gilles-duboscq: It's about using a different javac than the one from the JDK.
I don't think it's used anywhere. Maybe we should just remove the support.

java/com.oracle.mxtool.compilerserver/src/com/oracle/mxtool/compilerserver/JavacDaemon.java
tkrodriguez: Oh I see.  The class is the same in 9 but you don't need to reference tools.jar.  It's part of the core.  So I could directly import it in the suite but I'd need to have it on the javac command line but since 9 doesn't have a tools.jar I don't think I can do it.  Unless there's some way to conditionalize the deps?

mx.py
tkrodriguez: The comment isn’t visible in the PR for some reason.  Removing it would simplify things.  What alternate javac is there besides ecj?

mx.py
tkrodriguez: Your comment is missing from the PR for some reason.  I didn’t know about jdkLibrary.  I try to directly import it.

java/com.oracle.mxtool.compilerserver/src/com/oracle/mxtool/compilerserver/JavacDaemon.java
dougxc: Could you use `JavaCompiler`:
```
javax.tools.JavaCompiler jc = javax.tools.ToolProvider.getSystemJavaCompiler();
```

mx.py
dougxc: [error-prone](https://github.com/google/error-prone)

java/com.oracle.mxtool.compilerserver/src/com/oracle/mxtool/compilerserver/JavacDaemon.java
tkrodriguez: I've never seen that before.  It looks like the API is much higher level so it wouldn't integrate that easily.
<pre>   CompilationTask getTask(Writer out,
                            JavaFileManager fileManager,
                            DiagnosticListener<? super JavaFileObject> diagnosticListener,
                            Iterable<String> options,
                            Iterable<String> classes,
                            Iterable<? extends JavaFileObject> compilationUnits);<pre>

tkrodriguez: It does seem to solve the tools.jar issue tough.  You don't need to have tools.jar on your classpath for it to work.  I think it goes out and finds it.  Maybe it's not so hard to use.  I'm a little confused by the classes argument.  It looks like I need to tell it what classes need annotation processing.

tkrodriguez: It looks like it only supports a subset of the javac options and you have to do a lot of processing yourself to set up the arguments.  I don't think it helps.

dougxc: Ok - thanks for looking into it.

gilles-duboscq:
Regarding `tools.jar`, let's leave this as it is for now. I can look into using `jdkLibraries` later. it will probably need an update for JDK9 anyway.

tkrodriguez:
I've updated the comments and added utf8 handling on both sides.  https://travis-ci.org/tkrodriguez/graal-core/builds/116026507 is a graal-core build that's using my branch of mx.  The BuildJavaWithJavac times are about 1 minute instead of the normal 3.  If we could do the same for ecj that would help too I imagine.

mx.py
```diff
@@ -1897,11 +1919,97 @@ def buildJavacLike(self, jdk, project, jvmArgs, javacArgs, disableApiRestriction
...
+        JavacCompiler.__init__(self, None, extraJavacArgs)
+
+    def name(self):
+        return 'javac-daemon'
+
+    def run(self, jdk, jvmArgs, javacArgs):
+        # ignore the jvmArgs
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        s.connect(('127.0.0.1', self.daemon.port))
+        s.send(" ".join(javacArgs) + "\n".encode('utf-8'))
```
woess: you're only encoding `"\n"` here

dougxc:
I pulled this to my machines and saw compilation for `graal-core` and all its dependencies go from 115s to 42s - nice! I have a few final suggestions:

- add support for daemon in non-parallel mode (helps Windows which doesn't support parallelized compilation)
- generalize `--no-javac-daemon` option to `--no-daemon` in anticipation of `ECJDaemonCompiler`
- extra logging for daemon activity

Here's a patch implementing the above suggestions:

```
diff --git a/mx.py b/mx.py
index 9e8d575..31ca933 100755
--- a/mx.py
+++ b/mx.py
@@ -1726,7 +1726,7 @@ class JavaBuildTask(ProjectBuildTask):
         if self.args.jdt and not self.args.force_javac:
             return ECJCompiler(self.args.jdt, self.args.extra_javac_args)
         else:
-            if self.args.no_javac_daemon or self.args.alt_javac:
+            if self.args.no_daemon or self.args.alt_javac:
                 return JavacCompiler(self.args.alt_javac, self.args.extra_javac_args)
             else:
                 return JavacDaemonCompiler(self.args.extra_javac_args)
@@ -1938,7 +1938,9 @@ class JavacDaemonCompiler(JavacCompiler):
         # ignore the jvmArgs
         s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         s.connect(('127.0.0.1', self.daemon.port))
-        s.send(" ".join(javacArgs) + "\n".encode('utf-8'))
+        javacCommandLine = " ".join(javacArgs)
+        logv(self.daemon.jdk.javac + ' ' + javacCommandLine)
+        s.send(javacCommandLine + "\n".encode('utf-8'))
         f = s.makefile()
         retcode = int(f.readline().decode('utf-8'))
         s.close()
@@ -1967,7 +1969,7 @@ class JavacDaemon(Daemon):
     def __init__(self, jdk):
         self.jdk = jdk
         self.mainClass = 'com.oracle.mxtool.compilerserver.JavacDaemon'
-        build(['--no-javac-daemon', '--dependencies', 'com.oracle.mxtool.compilerserver'])
+        build(['--no-daemon', '--dependencies', 'com.oracle.mxtool.compilerserver'])
         self.coreCp = classpath(['com.oracle.mxtool.compilerserver'])
         # allocate a new port
         serversocket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
@@ -1986,10 +1988,12 @@ class JavacDaemon(Daemon):
             self.connection = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
             try:
                 self.connection.connect(('127.0.0.1', self.port))
+                logv('[Started ' + str(self) + ']')
                 return
             except socket.error as e:
                 retries = retries + 1
                 if retries > 5:
+                    logv('[Error starting ' + str(self) + ': ' + str(e) + ']')
                     raise e
                 else:
                     time.sleep(2)
@@ -2002,13 +2006,17 @@ class JavacDaemon(Daemon):
             try:
                 self.connection.send("\n")
                 self.connection.close()
+                logv('[Stopped ' + str(self) + ']')
             except socket.error as e:
-                logv('Error shutting down javac daemon: ' + str(e))
+                logv('Error stopping ' + str(self) + ': ' + str(e))

             self.proc.join()
             _removeSubprocess(self.sub)
             self.proc = None
             self.sub = None
+
+    def __str__(self):
+        return 'javac-daemon on port ' + str(self.port) + ' for ' + str(self.jdk)

 class ECJCompiler(JavacLikeCompiler):
     def __init__(self, jdtJar, extraJavacArgs=None):
@@ -8231,7 +8239,7 @@ def build(args, parser=None):
     parser.add_argument('--jdt-show-task-tags', action='store_true', help='show task tags as Eclipse batch compiler warnings')
     parser.add_argument('--alt-javac', dest='alt_javac', help='path to alternative javac executable', metavar='<path>')
     parser.add_argument('-A', dest='extra_javac_args', action='append', help='pass <flag> directly to Java source compiler', metavar='<flag>', default=[])
-    parser.add_argument('--no-javac-daemon', action='store_true', dest='no_javac_daemon', help='Disable use of daemon version of javac')
+    parser.add_argument('--no-daemon', action='store_true', dest='no_daemon', help='disable use of daemon Java compiler (if available)')

     compilerSelect = parser.add_mutually_exclusive_group()
     compilerSelect.add_argument('--error-prone', dest='error_prone', help='path to error-prone.jar', metavar='<path>')
@@ -8315,6 +8323,7 @@ def build(args, parser=None):
                 log(str(task))
         log("-- Serialized build plan --")

+    daemons = {}
     if args.parallelize:
         def joinTasks(tasks):
             failed = []
@@ -8366,7 +8375,6 @@ def build(args, parser=None):
                 cpus += t.parallelism
             return cpus

-        daemons = {}
         while len(worklist) != 0:
             while True:
                 active, failed = checkTasks(active)
@@ -8410,8 +8418,6 @@ def build(args, parser=None):
             worklist = sortWorklist(worklist)

         failed += joinTasks(active)
-        for daemon in daemons.values():
-            daemon.shutdown()

         if len(failed):
             for t in failed:
@@ -8420,8 +8426,12 @@ def build(args, parser=None):

     else:  # not parallelize
         for t in sortedTasks:
+            t.prepare(daemons)
             t.execute()

+    for daemon in daemons.values():
+        daemon.shutdown()
+
     # TODO check for distributions overlap (while loading suites?)

     if suppliedParser:
```

mx.py
tkrodriguez: oops thanks.

zapster:
@tkrodriguez one minor thing I noticed: it seems that the compileserver compilation does not honor the `--force-javac` flag:
```
mx build --force-javac
Compiling com.oracle.mxtool.compilerserver with JDT... [/home/zapster/graal/repos/mx.git/mxbuild/java/com.oracle.mxtool.compilerserver/bin/com/oracle/mxtool/compilerserver/ECJDaemon.class does not exist]
Compiling com.oracle.truffle.api with javac-daemon... [/home/zapster/graal/mx-workspaces/graal.git/truffle/mxbuild/truffle/com.oracle.truffle.api/bin/com/oracle/truffle/api/CompilerDirectives.class does not exist]
Compiling jdk.vm.ci.meta with javac-daemon... [/home/zapster/graal/mx-workspaces/graal.git/jvmci/mxbuild/jvmci/jdk.vm.ci.meta/bin/jdk/vm/ci/meta/RawConstant.class does not exist]
...
```
I guess this is not a big issue, I just wanted to mention it.

tkrodriguez:
It should probably respect all the command line args for completeness.  I'll pass the args in so we can do this.

tkrodriguez:
#104

Issue #51, opened 2016-02-22
Pass HTTP_PROXY to maven commands in mx


chumer:
Maven does not respect the HTTP_PROXY environment variable by default. So I think we should pass the proxy directly on each maven call in mx like this.
mvn -DproxySet=true -DproxyHost=myproxy.com -DproxyPort=3128 eclipse:eclipse

dougxc:
Sounds reasonable. @gilles-duboscq can you see any reason not to make this change?

gilles-duboscq:
Sounds good to me. We should just confirm that those arguments really work for maven.

dougxc:
@chumer can you please test the change and create a PR for it.

ldaynes:
can you clarify how one pass the proxy information to mx ? You need to pass both the proxy address and the port number separately right ? Are you planning to have some default mx env variable for indicating those ?

gilles-duboscq:
mx can extract the proxy data from the `http_proxy` environment variable.

dougxc:
We should pull the proxy env var to system property logic from [URLConnectionDownload](https://github.com/graalvm/mx/blob/master/URLConnectionDownload.java#L44) into mx.py to unify this for both maven and URL downloading.

dougxc:
This is fixed by #73.

PR #73, opened 2016-03-09
replaced Java download helper with pure python

dougxc:
@ldaynes, this also addresses #51.


mx.py
```diff
@@ -2096,15 +2096,15 @@ def _get_path_in_cache(name, sha1, urls, ext=None):
...
-    _, binDir = _compile_mx_class('URLConnectionDownload')
-    command = [get_jdk(tag=DEFAULT_JDK_TAG).java, '-cp', _cygpathU2W(binDir), 'URLConnectionDownload']
-    command += ['--no-progress', '--skip-download']
-    command += urls
-    def devnull(s):
-        pass
-    return run(command, err=devnull, nonZeroIsFatal=False) == 0
+    for url in urls:
+        try:
+            urllib2.urlopen(url, timeout=0.5).close()
```
gilles-duboscq: 0.5s timeout is rather short, do we really need an explicit timeout here?

dougxc: It was 0.3 before this change. In any case I think an explicit timeout is better than the default which seems to be very long.

gilles-duboscq: Ok, but shouldn't it be the same timeouts as the others (10)?

olpaw: I use download_file_exists to probe if the maven artifacts are already existing. So if the probing takes longer than the writing (in the common case) then it kinda defeats the purpose.

gilles-duboscq: ok

dougxc: Is that an accept then?

gilles-duboscq: yes

Issue #75, opened 2016-03-10
Allow to specify output directory of mx build


juergenchrist:
We would like to specify the directory where mx build should put files.  On larger systems this could be used to place build results on a ramdisk to speed up things.  At the moment I cannot specify the root directory, only the build subdirectory for every suite (as far as I see/understand).  Please add a command line option to specify the build root directory.

dougxc:
Fixed by #79

PR #79, opened 2016-03-11
added support for MX_ALT_OUTPUT_ROOT

dougxc:
This addresses #75.

Note that an mx global option cannot be used for this as some suites want their output root before command line options have been parsed.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -4757,6 +4757,8 @@ class Suite:
             outputRoot = suiteDict.get('outputRoot')
             if outputRoot:
                 self._outputRoot = os.path.realpath(_make_absolute(outputRoot.replace('/', os.sep), self.dir))
+            elif get_env('MX_ALT_OUTPUT_ROOT') is not None:
+                self._outputRoot = join(get_env('MX_ALT_OUTPUT_ROOT'), self.name)
             else:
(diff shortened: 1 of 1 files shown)


zapster:
looks good to me :+1:

mx.py
```diff
@@ -9567,17 +9575,36 @@ def processDep(dep, edge):
...
     # Changes to these values should cause regeneration of the project files.
-    value = 'logToConsole=%s\n' % logToConsole
-    if os.environ.get('PATH'):
-        value = value + 'PATH=%s\n' % os.environ['PATH']
-    if os.environ.get('JAVA_HOME'):
-        value = value + 'JAVA_HOME=%s\n' % os.environ['JAVA_HOME']
-    if os.environ.get('DEFAULT_VM'):
-        value = value + 'DEFAULT_VM=%s\n' % os.environ['DEFAULT_VM']
-    return value
+    settings = 'logToConsole=%s\n' % logToConsole
```
dougxc: @tkrodriguez , please check that this is semantically equivalent to how it was before my change.

juergenchrist:
Looks good to me.

mx.py
tkrodriguez: looks good

PR #81, opened 2016-03-12
GitConfig.update should check that the pull actually contains the requested revision

tkrodriguez:
Unlike mercurial you can't pull specific revisions you can only fetch and check for what you wanted.  When updating we should verify that we produced the proper revision and complain.  Otherwise you just end up with other mysterious failures because you're talking about a changeset that doesn't exist.

I tested this by putting a bogus revision in my suite.py and it's better than the old fatal bad object stuff.

Another side complaint here is that abortOnError doesn't default to true in many of these routines.  Is there any reason for this?

cc/ @dougxc

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -2730,7 +2730,7 @@ class VC(object):
         """
         abort(self.kind + ": outgoing is not implemented")
 
-    def push(self, vcdir, dest=None, rev=None, abortOnError=True):
+    def push(self, vcdir, dest=None, rev=None, abortOnError=False):
         """
         Push :param:`vcdir` at rev :param:`rev` to default if :param:`dest`
         is None, else push to :param:`dest`.
@@ -2756,7 +2756,7 @@ class VC(object):
         msg += ' with ' + self.proper_name
         log(msg)
 
-    def update(self, vcdir, rev=None, mayPull=False, clean=False, abortOnError=True):
+    def update(self, vcdir, rev=None, mayPull=False, clean=False, abortOnError=False):
         """
         update the :param:`vcdir` working directory.
         If :param:`rev` is not specified, update to the tip of the current branch.
@@ -3595,6 +3595,8 @@ class GitConfig(VC):
         """
         if rev and mayPull and not self.exists(vcdir, rev):
             self.pull(vcdir, rev=rev, update=False, abortOnError=abortOnError)
+            if not self.exists(vcdir, rev):
+                abort('Fetch of %s succeeded\nbut did not contain requested revision %s.\nCheck that the suite.py repository location is mentioned by \'git remote -v\'' % (vcdir, rev))
         cmd = ['git', 'checkout']
         if rev:
             cmd.extend(['--detach', rev])


zapster:
Thanks @tkrodriguez, looks good!

mx.py
```diff
@@ -3590,6 +3590,8 @@ def update(self, vcdir, rev=None, mayPull=False, clean=False, abortOnError=False
         """
         if rev and mayPull and not self.exists(vcdir, rev):
             self.pull(vcdir, rev=rev, update=False, abortOnError=abortOnError)
+            if not self.exists(vcdir, rev):
+                abort('Fetch of %s succeeded\nbut did not contain requested revision %s.\nCheck that your remotes agree with the suite.py reposity location' % (vcdir, rev))
```
dougxc: We should do the remote check here by calling and parsing `git remote -v`.

dougxc:
Since `abortOnError=True` in `VC.update`, then it should also be this way for `GitConfig.update`. Can you please fix it as part of the PR.

mx.py
zapster: Do we really need to? What do we gain?

dougxc: It helps those who do not know what checking remotes against suite.py means or how to do it. Since mx is trying to abstract over VC operations in general, error messages specifically mentioning git concepts and actions can be confusing to the uninitiated. At the very least, the error message can show the git command for checking remotes as well as the full path to the suite.py to check.

zapster: Ok, makes sense.

woess: can't we just `git fetch` from the URLs in suite.py instead?

btw: s/reposity/repository/

tkrodriguez: Fixed the typo.  The only way to get into the state where your origin and the suite.py location are out of agreement is if you know something about using git.  So I think telling them to check their remotes using git remote -v should be sufficient, so I'l update the text a little bit.  The text will now read 
<pre>
$ mx sfor
Fetch of /Users/tkrodrig/ws/basic-graal/truffle succeeded
but did not contain requested revision e322868c76f775a399bf96915c95e077c4f2c08c.
Check that the suite.py repository location is mentioned by 'git remotes -v
</pre>

I don't think that we should do a direct fetch for them either.  git really wants you to do this through your configured remotes so I think we should treat this as a misconfiguration.  We also don't know the proper URL down inside the VC.

tkrodriguez:
I can fix the abortOnError but it's also false in the hg version of update and push.  Should I fix those as well?

mx.py
woess: Hm, yeah, I see your point regarding "git really wants you to do this through your configured remotes". We could probably attempt to detect the remote name by URL based on `git remote -v` output, but it's non-trivial, since it depends on the protocol used (https, ssh). So I'm fine with this.

dougxc:
Hmm, maybe that means the default value in the `VC` should be `False`. It's probably safer to fix that than the two existing implementations since that behavior may be relied upon already.

mx.py
dougxc: Ok, I buy your point of git knowledge and experience being assumed to even get in this situation so the change is good as is.

tkrodriguez:
I'm leery of changing it but at the same time the current value seems wrong.  I think I'll leave it alone and we can have a separate discussion about the proper setting.

dougxc:
Can at least then make the method in `VC` consistent with the implementations.

tkrodriguez:
I've made all implementations agree that for push and update abortOnError=False.  The only other minor inconsistency is that update=True for git.pull but False for all the others.  It sort of looks like all paths explicitly pass update anyway so that may not matter.

Issue #71, opened 2016-03-08
Provide def release_version(self) for BinarySuite


olpaw:
The call to _updateGraalPropertiesFile()  in `mx.graal-core/mx_graal_8.py line 73` will fail with jvmci imported as binary suite because there is no definition of def release_version(self) for BinarySuite. A proper definition needs to be added.

dougxc:
Have you seen a failure? If so, under what conditions?

This should not fail as this code is only called when building the JVMCI JDK and when JVMCI is a binary suite this step is not necessary.

olpaw:
> Have you seen a failure? If so, under what conditions?

Yes. We have a project that transitively depends on jvmci that builds (and uses) jvm-product-server-linux-amd64. If I binary deploy all transitive dependencies and then use them via `MX_BINARY_SUITES='' mx build` it will work fine except for binary artifact for jvm-product-server-linux-amd64. There mx tries to call `_updateGraalPropertiesFile()` and fails in line 73 because now jvm-product-server-linux-amd64 comes from a binary suite (which does not have `def release_version(self) defined`)

But when I define a dummy implementation for release_version(self) in BinarySuite everything works a expected.

dougxc:
When you execute `MX_BINARY_SUITES='' mx build` isn't the jvmci source suite resolved (and downloaded if necessary) bypassing the jvmci binary suite altogether?

olpaw:
I will send you a console log via email.

dougxc:
Sorry, I got confused and interpreted `MX_BINARY_SUITES=‘’` to mean don’t use binary suites. Also, I didn’t see any failure with graal-core as the primary suite. Obviously, it has to be an imported binary suite to see the problem!

What does your dummy implementation look like? Worth creating a PR from it?

olpaw:
> Worth creating a PR from it?

I don't think so:
```
def release_version(self):
    return self.version()
```

dougxc:
That seems reasonable to me - where else could we get version info from?

PR #82, opened 2016-03-14
Provide release_version for class BinarySuite

olpaw:
In mx.graal-core/mx_graal_8.py _updateGraalPropertiesFile() calls
release_version for the jvmci suite. In the binary suite usecase this
results in: AttributeError: BinarySuite instance has no attribute
'release_version'. To make jvmci usable as binary suite release_version
is provided for class BinarySuite.

See: https://github.com/graalvm/mx/issues/71


dougxc:
This is actually more about making graal-core available as a binary suite.

PR #85, opened 2016-03-16
Add JMH and `mx microbench` command

### LINKED ISSUE #84: Move `mx microbench` from graalvm/graal-core to mx
In graalvm/graal-core we have the `mx microbench` [command](https://github.com/graalvm/graal-core/blob/9e1f376e6dac74cab81f57fb98396366e896f4e1/mx.graal-core/mx_graal_8.py#L111)  which looks for [JMH](http://openjdk.java.net/projects/code-tools/jmh/) benchmarks and runs them. This functionality should be moved to mx.

zapster:
Fixes #84. This change introduces the `mx microbench` command as known from graal-core.

The core functionality is implemented in `MicrobenchExecutor` which can be sub-classed by downstream projects and attached using `mx_microbench.set_microbench_executor`.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -12838,7 +12838,7 @@ def main():
         # no need to show the stack trace when the user presses CTRL-C
         abort(1)
 
-version = VersionSpec("5.13.0")
+version = VersionSpec("5.14.0")
 
 currentUmask = None
 
--- a/mx.py
+++ b/mx.py
@@ -67,6 +67,7 @@ import mx_findbugs
 import mx_sigtest
 import mx_gate
 import mx_compat
+import mx_microbench
 
 ERROR_TIMEOUT = 0x700000000 # not 32 bits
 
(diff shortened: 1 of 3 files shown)


zapster:
testing done here: graalvm/graal-core#133

mx.py
```diff
@@ -12836,7 +12838,7 @@ def alarm_handler(signum, frame):
         # no need to show the stack trace when the user presses CTRL-C
         abort(1)
 
-version = VersionSpec("5.13.0")
+version = VersionSpec("5.13.1")
```
zapster: As this adds only new functionality, this is only a bugfix bump, right?

mx_microbench.py
```diff
@@ -0,0 +1,112 @@
...
+# ----------------------------------------------------------------------------------------------------
+
+import mx
+from argparse import ArgumentParser
+
+_microbench_executor = None
+
+def set_microbenchmark_executor(ex):
+    global _microbench_executor
+    assert _microbench_executor is None, 'cannot override microbenchmark executor twice'
```
zapster: It two downstream projects, say `graal-core` and `truffle`, want to change the executor we have a problem. But the same is true for the unittest vm launcher. I guess we are fine for the moment. If this turns out to be a problem we can change it.

mx.py
dougxc: No, you need a minor version bump. From `README.md`:

```
- If you add new functionality without breaking backward compatibility, leave
the major as it is, increase the minor number and reset the patch level.
```

mx_microbench.py
dougxc: Ok.

PR #86, opened 2016-03-17
Split out daemon logic

tkrodriguez:
This generalizes the daemon support to add ECJ.  ecj parses the command line so I had to perform some extra tricks to maintain the separators between arguments.  The ECJ build times went from 2:42 to 1:10 in travis.  Locally it saved about 40 seconds on my mac.

@gilles-duboscq @woess

### DIFF
--- a/java/com.oracle.mxtool.compilerserver/src/com/oracle/mxtool/compilerserver/ECJDaemon.java
+++ b/java/com.oracle.mxtool.compilerserver/src/com/oracle/mxtool/compilerserver/ECJDaemon.java
@@ -22,32 +22,38 @@
  */
 package com.oracle.mxtool.compilerserver;
 
+import java.io.PrintWriter;
 import java.lang.reflect.Method;
 
-public class JavacDaemon extends CompilerDaemon {
+public class ECJDaemon extends CompilerDaemon {
 
-    private final class JavacCompiler implements Compiler {
+    private final class ECJCompiler implements Compiler {
         public int compile(String[] args) throws Exception {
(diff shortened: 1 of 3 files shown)


java/com.oracle.mxtool.compilerserver/src/com/oracle/mxtool/compilerserver/ECJDaemon.java
```diff
@@ -0,0 +1,59 @@
...
+import java.lang.reflect.Method;
+
+public class ECJDaemon extends CompilerDaemon {
+
+    private final class ECJCompiler implements Compiler {
+        public int compile(String[] args) throws Exception {
+            for (int i = 0; i < args.length; i++) {
+                if (args[i].length() == 0) {
+                    args[i] = "\"\"";
+                }
```
gilles-duboscq: Do we also need to quote arguments that contain spaces?
What about arguments that contain quotes?

dougxc: Why don't we use this entry point:
```
public static boolean compile(java.lang.String[], java.io.PrintWriter, java.io.PrintWriter, org.eclipse.jdt.core.compiler.CompilationProgress);
```
That would avoid any need to massage the args.

tkrodriguez: Yes I missed that last night.  The documentation I found led me to the String one and I didn't do much searching for a better one.  I'll fix it.

PR #87, opened 2016-03-17
Make NativeTARDistribution usable

olpaw:
Fixes bugs/missing features that make NativeTARDistribution usable for binary deployment of non-java projects.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -1096,9 +1096,10 @@ Attributes:
     path: suite-local path to where the tar file will be placed
 """
 class NativeTARDistribution(Distribution):
-    def __init__(self, suite, name, deps, path, excludedLibs, platformDependent, theLicense):
+    def __init__(self, suite, name, deps, path, excludedLibs, platformDependent, theLicense, relpath):
         Distribution.__init__(self, suite, name, deps, excludedLibs, platformDependent, theLicense)
         self.path = _make_absolute(path, suite.dir)
+        self.relpath = relpath
 
     def make_archive(self):
         directory = dirname(self.path)
@@ -1108,8 +1109,13 @@ class NativeTARDistribution(Distribution):
             for d in self.archived_deps():
                 if not d.isNativeProject():
                     abort('Unsupported dependency for native distribution {}: {}'.format(self.name, d.name))
+                output = d.getOutput()
+                output = join(self.suite.dir, output) if output else None
                 for r in d.getResults():
-                    filename = basename(r)
+                    if output and self.relpath:
+                        filename = os.path.relpath(r, output)
+                    else:
+                        filename = basename(r)
                     assert filename not in files, filename
                     # Make debug-info files optional for distribution
                     if is_debug_lib_file(r) and not os.path.exists(r):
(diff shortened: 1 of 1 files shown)


dougxc:
@gilles-duboscq , can you  please have a look at this

mx.py
```diff
@@ -1140,9 +1141,17 @@ def localExtension(self):
...
+        tarfilename = None
         with gzip.open(f, 'rb') as gz, open(f[:-len('.gz')], 'wb') as tar:
             shutil.copyfileobj(gz, tar)
+            tarfilename = tar.name
         os.remove(f)
+        if self.output:
+            output = join(self.suite.dir, self.output)
+            assert tarfilename
+            with tarfile.open(tarfilename, 'r:') as tar:
+                log('Extract {} to {}'.format(tarfilename, output))
```
gilles-duboscq: maybe `s/Extract/Extracting/`?

gilles-duboscq:
So the `output` argument only has a meaning when the distribution comes from a `BinarySuite`?

This feels a bit strange, how does it work when the distribution comes from a `SourceSuite`?
It looks like extracting to `output` should be both a post-pull and post-build action.

olpaw:
@gilles-duboscq only if the distribution is native the output argument is considered. And native distributions can only contain native projects. Here is the usecase:
```
"XXX_HOSTED_NATIVE" : {
  "dependencies" : ["com.oracle.xxx.native"],
  "native" : True,
  "relpath" : True,
  "output" : "clibraries",
  "os_arch" : {
    "linux" : {
      "amd64" : {
          "path" : "build/xxx-hosted-native-linux.tar",
      },
    },
    "darwin" : {
      "amd64" : {
        "path" : "build/xxx-hosted-native-darwin.tar",
      }
    },
  }
}, 
```
Specifying `"output" : "clibraries"` in the above native distribution specification ensures that after the tar-ball got pulled it will extract its contents to output (relative to the imported binary suite dir).
Specifying `"relpath" : True` ensures that sub-directory structures are replicated into the tar-ball on creation (e.g. libfoo.so, include/foo.h ... here it is important that we have include/foo.h in the tarball instead of foo.h (what basename would give us)).

mx.py
olpaw: Fixed.

olpaw:
Would it help if I rename the `output` attr to `extract`?

olpaw:
> This feels a bit strange, how does it work when the distribution comes from a SourceSuite?
It looks like extracting to output should be both a post-pull and post-build action.

If you are in SourceSuite mode output has no effect. It's only relevant for the BinarySuite case to reconstruct certain folders that the suite depends on being at specific places. I thought it would be convenient for other users as well (instead of having each user reinvent the wheel with a post-pull action). BTW, how would I hook into post-pull for a given suite anyway?

gilles-duboscq:
So the `output` attribute has to match exactly the `output` attribute of the project that was included in the tar. We could also so that automatically. However that's not completely straightforward because project data is not currently loaded for `BinarySuites`.
I created #88 to track that.

PR #93, opened 2016-03-17
support project javaCompliance value that ends with "+" as a lower bound

dougxc:
Fixes this issue:
```
SystemExit: dependency named com.oracle.mxtool.compilerserver was not found:
  project com.oracle.mxtool.compilerserver was removed as Java compliance 1.7 cannot be satisfied by configured JDKs
```


tkrodriguez:
looks good

gilles-duboscq:
Is it something like "higher is better even if `--strict-compliance` is enabled"?

dougxc:
No, it currently selects the first compatible JDK. I don't see a need for anything more specific than that.

PR #94, opened 2016-03-18
Allow user to contol how sforceimports handles a dirty repository

tkrodriguez:
scheckimports has an option to not complain if there are uncommitted changes and sforceimports should have the same support.  It's pretty benign in scheckimports but for sforceimports git or hg will attempt to merge so it will still ask if you want to continue.  So for example:

<pre>
$ mx sforce -i
WARNING: Uncommited changes in jvmci. Attempt to update anyway? [Yn]:
merging src/share/vm/jvmci/jvmciRuntime.cpp
merging src/share/vm/jvmci/jvmciRuntime.hpp
27 files updated, 2 files merged, 1 files removed, 0 files unresolved
</pre>

@dougxc

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -2954,6 +2954,17 @@ class VC(object):
         """
         abort(self.kind + " isDirty is not implemented")
 
+    def status(self, vcdir, abortOnError=True):
+        """
+        report the status of the repository
+
+        :param str vcdir: a valid repository path
+        :param bool abortOnError: if True abort on mx error
+        :return: True on success, False otherwise
+        :rtype: bool
+        """
+        abort(self.kind + " status is not implemented")
+
     def locate(self, vcdir, patterns=None, abortOnError=True):
         """
         Return a list of paths under vc control that match :param:`patterns`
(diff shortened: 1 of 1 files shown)


gilles-duboscq:
Could we be a little bit more descriptive about what's going to happen?
e.g., could this erase my local changes? will the VC attempt a "safe" merge? which VC are we talking about?

tkrodriguez:
Both hg and git attempt to merge outstanding changes when performing an update.  You can lose changes if you pull in something equivalent to on of  your local edits.  So for example if you backport a changeset to an earlier point and then move to a more recent one it will disappear.  Switching back to that earlier point will have lost that uncommitted backport.  I was just hitting this last night when investigating the jvmci.Compiler property problem.

Maybe it should be --merge-uncommitted to make it clearer what's expected to happen?  Any suggestions for the message itself would be appreciated.

tkrodriguez:
Also maybe it should give you the choice of truly cleaning out the changes, attempting to merge or aborting if it notices that it's dirty.  That would be better I think.  I'll rework this.

tkrodriguez:
I've pushed a few version that gives a few options for handling.  So you get something like this:

<pre>
$ mx sfor
WARNING: There are uncommited changes in jvmci! (a)bort, (s)how, (c)lean, (m)erge?? [Ascm]: s
M jvmci/jdk.vm.ci.hotspot/src/jdk/vm/ci/hotspot/HotSpotResolvedObjectTypeImpl.java
M jvmci/jdk.vm.ci.runtime.test/src/jdk/vm/ci/runtime/test/TypeUniverse.java
? make/closed
? src/closed
WARNING: There are uncommited changes in jvmci! (a)bort, (s)how, (c)lean, (m)erge?? [Ascm]: m
merging jvmci/jdk.vm.ci.hotspot/src/jdk/vm/ci/hotspot/HotSpotResolvedObjectTypeImpl.java
4 files updated, 1 files merged, 0 files removed, 0 files unresolved
</pre>

I added a new status command to the VCs and generalized ask_yes_no as well.  cc @dougxc

PR #95, opened 2016-03-18
Only forward SIGQUIT to the child Java process not the whole process group

tkrodriguez:
mx tries to forward SIGQUIT to its children but currently it signals the whole process group which appears to contain mx itself, resulting a storm of SIGQUITs instead of a single SIGQUIT.

@gilles-duboscq I think this will let us make the buildbot timeout code do a few dumps spaced over time so we can see what's happening.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -8087,7 +8087,8 @@ def _send_sigquit():
             if get_os() == 'windows':
                 log("mx: implement me! want to send SIGQUIT to my child process")
             else:
-                _kill_process_group(p.pid, sig=signal.SIGQUIT)
+                # only send SIGQUIT to the child not the process group
+                os.kill(p.pid, signal.SIGQUIT)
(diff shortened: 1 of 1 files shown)


dougxc:
This PR seems to include https://github.com/graalvm/mx/pull/94 which I'm sure you'll sort out before integrating.

tkrodriguez:
Oops branched from the wrong point.  Fixed.

tkrodriguez:
So this works fine from the command line but our timed out runs in buildbot no longer have stack dumps.  https://lafo.ssw.uni-linz.ac.at/buildbot/builders/CI-graal-core_gate_amd64_linux_postgate_jdk9-SNAPSHOT-fastdebug_jdk8_jdk7/builds/97/steps/mxgate/logs/stdio  Not sure why yet.

PR #96, opened 2016-03-18
enable use of Java compiler daemon on Windows

dougxc:
@wirthi can you please test this on your machine.


tkrodriguez:
looks good

wirthi:
Thanks, building now works again on windows. "Compiling com.oracle.truffle.api with javac-daemon...". It feels maybe a bit faster, but I did not measure it (before).

PR #97, opened 2016-03-21
disable Java compilation daemon when building from NetBeans

dougxc:
NetBeans does not benefit from daemon-based compilation so disable it.


jtulach:
The compilation from NetBeans works correctly now. Thanks.

PR #98, opened 2016-03-21
Java compiler daemon must be in its own process group

dougxc:
Without isolating the daemon in its own process group, killing it during `mx.abort()` kills the process group of mx which is problematic when called from NetBeans.


tkrodriguez:
Is this something that multiprocessing used to do for us?

gilles-duboscq:
The other "compilers" in mx use `run` which takes care of that.
In the initial review i wondered about why you are not using `run` but i guess it's because we would need to add some "async" option to it.

dougxc:
Yes, I took this option in lieu of adding an `async` argument to `run`.

PR #99, opened 2016-03-21
more robust handling of long pathnames on Windows

dougxc:
This adds better support for long pathnames on Windows for operations in the Python `os` and `os.path` modules that otherwise don't support them ([Issue1899](https://bugs.python.org/issue18199)).

I'm not sure if intercepting the builtin `open` function is the right way to go. The alternative is to refactor all uses of this function in mx (and possibly mx extensions) to use the `_safe_path` function added by this PR.


dougxc:
This may help you @wirthi

gilles-duboscq:
:+1:

PR #100, opened 2016-03-22
make inter-thread control field volatile

dougxc:
/cc @gilles-duboscq

### DIFF
--- a/java/com.oracle.mxtool.compilerserver/src/com/oracle/mxtool/compilerserver/CompilerDaemon.java
+++ b/java/com.oracle.mxtool.compilerserver/src/com/oracle/mxtool/compilerserver/CompilerDaemon.java
@@ -145,11 +145,6 @@ public abstract class CompilerDaemon {
                         logf("Result = %d\n", result);
 
                         output.write(result + "\n");
-
-                        // Try close open file handles (may help on Windows)
-                        System.gc();
-                        System.gc();
-                        System.runFinalization();
                     }
                 } finally {
                     // close IO streams, then socket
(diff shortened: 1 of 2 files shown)


tkrodriguez:
Why is it safe to remove the exit call?  The main thread is asleep in accept() so won't it sleep there forever?  We could explicitly close the serverSocket and remove the awaitTermination logic and just the let the whole thing exit once there are no daemon threads left but I put in the exit the ensure termination.

dougxc:
Ok, then I don't think we need `running` at all right?

tkrodriguez:
If we explicitly close the socket then we don't need the running flag, though we have to handle the exception that accept will throw and there's no reason to report that exception if it's an orderly shutdown so we might need a flag anyway.

I'm not entirely comfortable with removing the exit though.  I like the idea that someone is explicitly shutting down the daemon, not just relying on it exiting when there are no daemons left.

dougxc:
Ok, let's just make `running` volatile and leave it at that.

tkrodriguez:
I'm ok with that.

java/com.oracle.mxtool.compilerserver/src/com/oracle/mxtool/compilerserver/CompilerDaemon.java
```diff
@@ -134,11 +145,6 @@ public void run() {
                         logf("Result = %d\n", result);
 
                         output.write(result + "\n");
-
-                        // Try close open file handles (may help on Windows)
-                        System.gc();
-                        System.gc();
-                        System.runFinalization();
```
dougxc: These statements should never have been added in the first place. The problem they were aimed at has a completely different cause (see #99) and they have been observed to cause the daemon to slow down significantly on Linux.

PR #102, opened 2016-03-23
Add TestNG support.

christhalinger:
SQE is adding some new JVMCI tests and is using TestNG.  I want them to integrate the new tests with mx.


dougxc:
This is just adding the TestNG library to mx without any command for running TestNG tests. It would make more sense to add the library to the mx suite that includes TestNG based sources and (possibly) has an mx extension command for running them.

christhalinger:
Yeah, I noticed that as well after I was poking around in the new tests.  ``unittest`` should support running TestNG tests.  I'm looking into it right now.

dougxc:
Christian, are you still looking into modifying the `unittest` command or can I close this issue as won't fix?

christhalinger:
I was wondering about this myself :)  At this point I don't care, so we can close it.

PR #104, opened 2016-03-24
Compilation of compilerserver should respect --force-javac

tkrodriguez:
I think --force-javac is the only flag it really needs to respect or are there others?

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -1971,9 +1971,11 @@ class Daemon:
         pass
 
 class CompilerDaemon(Daemon):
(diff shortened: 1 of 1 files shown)


zapster:
looks good, thanks!

zapster:
@tkrodriguez maybe `--warning-as-error`?

tkrodriguez:
Well problems in the mx compilerserver shouldn't cause my build to fail and accepting arbitrary flags seems to open us up to that.  cc @dougxc I'm even leery of --warning-as-error for that reason.

PR #106, opened 2016-03-24
Add `mx_benchmark.py`.

axel22:
Review by @gilles-duboscq

### DIFF
--- a/mx_benchmark.py
+++ b/mx_benchmark.py
@@ -230,7 +230,7 @@ class StdOutBenchmarkSuite(BenchmarkSuite):
         raise NotImplementedError()
 
 
-class JavaBenchmarkSuite(StdOutBenchmarkSuite): #pylint: disable=R0921
+class JavaBenchmarkSuite(StdOutBenchmarkSuite): #pylint: disable=R0922
     """Convenience suite used for benchmarks running on the JDK.
     """
     def defaultVmArgs(self):
--- a/mx_benchmark.py
+++ b/mx_benchmark.py
@@ -1,0 +1,88 @@
+#
+# ----------------------------------------------------------------------------------------------------
+#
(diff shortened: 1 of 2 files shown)


mx_benchmark.py
```diff
@@ -0,0 +1,304 @@
...
+        Arguments:
+            benchmarks (list of strings): List of benchmark names.
+            vmargs (list of strings): List of VM arguments.
+            runargs (list of strings): List of benchmark arguments.
+
+        Returns:
+            object: A list of measurement results.
+
+            A measurement result is an object that can be converted into JSON and is
+            merged with root.
```
dougxc: What's `root`?

mx_benchmark.py
```diff
@@ -0,0 +1,304 @@
...
+                  scoring function (e.g. `"1.54"`)
+                - `metric.type` -- a string denoting the type of the metric
+                  (e.g. `"numeric"`)
+                - `metric.better` -- `"higher"` if higher is better, `"lower"` otherwise
+                - `metric.iteration` -- iteration number of the measurement (e.g. `0`)
+        """
+        raise NotImplementedError()
+
+
+def add_benchmark_suite(suite):
```
dougxc: Maybe we should adopt the naming convention of `bmsuite` for BenchmarkSuite typed variables to minimize the confusion with normal mx suites.

mx_benchmark.py
axel22: Meant as the root data point object. I'll rewrite it to just `merged with the data point object`.

mx_benchmark.py
axel22: Sounds good.

axel22:
@gilles-duboscq PTAL

PR #107, opened 2016-03-25
travis: use oraclejdk8.

zapster:
Since `com.oracle.mxtool.junit` currently wants java 8 we should also use it in travis.


zapster:
Alternatively we could also change the dependency to 1.7.

dougxc:
We should leave the compliance level at 1.8 since it's so easy to satisfy the dependency.

PR #109, opened 2016-03-25
Ignore mx.mx/eclipse-launches

tkrodriguez:
#103 causes eclipse-launches to be generated in mx so we should ignore those files  cc @mlvdv


tkrodriguez:
This was added by https://github.com/graalvm/mx/commit/765e8776

PR #110, opened 2016-03-28
eclipseinit skips NativeProject dependencies

### LINKED ISSUE #108: mx eclipseinit adds native projects to .classpath
mx eclipseinit should not add a project with "native : True" to the .classpath file.  This generates an invalid eclipse configuration.

juergenchrist:
We do not add non-existing source folders of NativeProjects to the
.classpath files generated by eclipseinit.

This fixes #108.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -9728,7 +9728,7 @@ def _eclipseinit_project(p, files=None, libFiles=None):
             libFiles.append(path)
 
     for dep in sorted(projectDeps):
-        if not isinstance(dep, NativeProject):
+        if not dep.isNativeProject():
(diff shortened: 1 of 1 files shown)


mx.py
```diff
@@ -9728,7 +9728,8 @@ def processDep(dep, edge):
             libFiles.append(path)
 
     for dep in sorted(projectDeps):
-        out.element('classpathentry', {'combineaccessrules' : 'false', 'exported' : 'true', 'kind' : 'src', 'path' : '/' + dep.name})
+        if not isinstance(dep, NativeProject):
```
dougxc: There are a bunch of type test methods defined on the `Dependency` class. Please change the above to use `isNativeProject()` instead.

PR #112, opened 2016-03-31
add --dynamicimports alias; remove bitbucket doc link

mickjordan:


mx.py
```diff
@@ -6809,7 +6808,7 @@ def __init__(self, parents=None):
         self.add_argument('--src-suitemodel', help='mechanism for locating imported suites', metavar='<arg>')
         self.add_argument('--dst-suitemodel', help='mechanism for placing cloned/pushed suites', metavar='<arg>')
         self.add_argument('--primary', action='store_true', help='limit command to primary suite')
-        self.add_argument('--dynamicimport', action='append', dest='dynamic_imports', help='dynamically import suite <name>', metavar='<name>', default=[])
+        self.add_argument('--dynamicimport', '--dynamicimports', action='append', dest='dynamic_imports', help='dynamically import suite <name>', metavar='<name>', default=[])
```
dougxc: I think you can just change `--dynamicimport` to `--dynamicimports` and still be backwards compatible as `ArgumentParser` does shortest unique prefix matching for options.

mickjordan: Ah, forgot that little feature. Will do

dougxc:
Merged internally.

PR #118, opened 2016-07-21
Support exploded JDK.

christhalinger:
Using an exploded JDK results in:

```
$ mx --java-home=/Users/cthalinger/jvmci/build/macosx-x86_64-normal-server-release/jdk eclipseinit
  File "/Users/cthalinger/jvmci/graal-core/mx.graal-core/suite.py", line 274 in definition of com.oracle.graal.api.replacements:
JDK library JVMCI_API required by com.oracle.graal.api.replacements not provided by Java 1.8.0_92-internal (1.8) from /Users/cthalinger/jvmci/build/macosx-x86_64-normal-server-release/jdk
```

This also needs a change in the graal-core suite file:

```
diff --git a/mx.graal-core/suite.py b/mx.graal-core/suite.py
index 364136c..bf5dbf6 100644
--- a/mx.graal-core/suite.py
+++ b/mx.graal-core/suite.py
@@ -21,14 +21,14 @@ suite = {
 
   "jdklibraries" : {
     "JVMCI_SERVICES" : {
-      "path" : "jre/lib/jvmci-services.jar",
-      "sourcePath" : "jre/lib/jvmci-services.src.zip",
+      "path" : "lib/jvmci-services.jar",
+      "sourcePath" : "lib/jvmci-services.src.zip",
       "optional" : False,
       "jdkStandardizedSince" : "9",
     },
     "JVMCI_API" : {
-      "path" : "jre/lib/jvmci/jvmci-api.jar",
-      "sourcePath" : "jre/lib/jvmci/jvmci-api.src.zip",
+      "path" : "lib/jvmci/jvmci-api.jar",
+      "sourcePath" : "lib/jvmci/jvmci-api.src.zip",
       "dependencies" : [
         "JVMCI_SERVICES",
       ],
@@ -36,8 +36,8 @@ suite = {
       "jdkStandardizedSince" : "9",
     },
     "JVMCI_HOTSPOT" : {
-      "path" : "jre/lib/jvmci/jvmci-hotspot.jar",
-      "sourcePath" : "jre/lib/jvmci/jvmci-hotspot.src.zip",
+      "path" : "lib/jvmci/jvmci-hotspot.jar",
+      "sourcePath" : "lib/jvmci/jvmci-hotspot.src.zip",
       "dependencies" : [
         "JVMCI_API",
       ],
```

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -2803,13 +2803,21 @@ class JdkLibrary(BaseLibrary, ClasspathDependency):
         else:
             return NotImplemented
 
+    def get_jdk_path(self, jdk, path):
+        # Exploded JDKs don't have a jre directory.
+        if exists(join(jdk.home, 'jre')):
+            return join(jdk.home, 'jre', path)
+        else:
+            return join(jdk.home, path)
+
     def is_provided_by(self, jdk):
         """
         Determines if this library is provided by `jdk`.
 
         :param JDKConfig jdk: the JDK to test
         """
-        return jdk.javaCompliance >= self.jdkStandardizedSince or exists(join(jdk.home, self.path))
+        path = self.get_jdk_path(jdk, self.path)
+        return jdk.javaCompliance >= self.jdkStandardizedSince or exists(path)
 
     def getBuildTask(self, args):
         return NoOpTask(self, args)
@@ -2827,7 +2835,7 @@ class JdkLibrary(BaseLibrary, ClasspathDependency):
             abort('A JDK is required to resolve ' + self.name)
         if jdk.javaCompliance >= self.jdkStandardizedSince:
             return None
-        path = join(jdk.home, self.path)
+        path = self.get_jdk_path(jdk, self.path)
         if not exists(path):
             abort(self.name + ' is not provided by ' + str(jdk))
         return path
@@ -2841,7 +2849,8 @@ class JdkLibrary(BaseLibrary, ClasspathDependency):
         """
         if self.sourcePath is None:
             return None
-        return self.sourcePath if isabs(self.sourcePath) else join(jdk.home, self.sourcePath)
+        path = self.get_jdk_path(jdk, self.sourcePath)
+        return self.sourcePath if isabs(self.sourcePath) else path
 
     def isJar(self):
         return True


dougxc:
Why is this necessary? Are you building a JDK8 with a non-standard layout? Or setting `JAVA_HOME` to the `jre/` subdirectory in a JDK?

christhalinger:
I'm using `mx --java-home=$(JDK_OUTPUTDIR)` because that's the only way to build Graal with the just built JDK.  I am integrating Graal into the JDK build process and all this has to happen before `make images`.

dougxc:
Ok, sounds reasonable. Can you please bump the patch level number of the `version` field down the bottom of `mx.py`.

christhalinger:
What happened here?  It was merged and then reverted?!?  The patch is even gone in my branch now...

sanzinger:
Sorry, it had broke our continuous integration builds with JDK 8. I had to take it out immediately. I'll provide a fix for this issues and re-merge.

christhalinger:
What about the version bump, as @dougxc requested?

sanzinger:
Yes, I've seen that later, but it is not related. I'm working right now to fix the issue.

christhalinger:
Ok, thanks.

mx.py
```diff
@@ -2803,13 +2803,21 @@ def __eq__(self, other):
         else:
             return NotImplemented
 
+    def get_jdk_path(self, jdk, path):
+        # Exploded JDKs don't have a jre directory.
+        if exists(join(jdk.home, 'jre')):
```
sanzinger: The issue here is, that mx.graal-core/suites.py links to the jre directory
```
    "JVMCI_API" : {
      "path" : "jre/lib/jvmci/jvmci-api.jar",
....
    },
```
in that case, following path is created: ``$JAVA_HOME/jre/jre/lib/jvmci/jvmci-api.jar``. We could do a bit of fuzzying like:

```
    def get_jdk_path(self, jdk, path):
        # Exploded JDKs don't have a jre directory.
        if exists(join(jdk.home, path)):
            return join(jdk.home, path)
        elif exists(join(jdk.home, 'jre')):
            return join(jdk.home, 'jre', path)
        else:
            return join(jdk.home, path)

```

this is not beautilful but in order to get this feature and have backwards compatibility we have to do it.
In the long run, we may remove the ``jre`` prefix from the suites.py dependencies.

christhalinger: Right.  I didn't think of being backwards compatible.

sanzinger:
Ok, I've prepeared an internal review with the second (simpler) version of ``get_jdk_path``. If this fits to you and I have an internal approval, I'll merge it.

christhalinger:
I haven't tried the patch but it looks ok.  Did you try it?

dougxc:
> I'm using mx --java-home=$(JDK_OUTPUTDIR) because that's the only way to build Graal with the just built JDK. I am integrating Graal into the JDK build process and all this has to happen before make images.

In this scenario, which JDK version are you building? Based on your mention of `make images` it sounds like JDK9 in which case the `"jdkStandardizedSince" : "9"` setting in `suite.py` should make the `"path"` attribute effectively ignored. Is the just built JDK somehow reporting the wrong version?

christhalinger:
No, it's 8u; it also has `make images`.

dougxc:
Ok, thanks for the clarification. I guess you're deploying `graal.jar` alongside the JVMCI jars as part of the `images` build step.

christhalinger:
Yes.  By doing that we have `graal.jar` in all the various JDK images (exploded, jre, jdk).

christhalinger:
Yeah, this doesn't work right now.  Are you guys pushing a change to `graal-core` to remove the `jre` prefix from the libraries?

dougxc:
Done (https://github.com/graalvm/graal-core/commit/1bc73f6843a48e4c18e06a1a0473fdba727c4cff)

christhalinger:
Thanks!

PR #119, opened 2016-08-24
Adding Jackpot 3.0 support

jlahoda:
Hi,

This patch adds a support for Jackpot 3.0 - Jackpot 3.0 interprets custom hint files, and reports warnings as appropriate. To use it, it should suffice to do "mx jackpot", similarly to "mx findbugs".

Comments are welcome!

Thanks!

### DIFF
--- a/mx.mx/suite.py
+++ b/mx.mx/suite.py
@@ -110,6 +110,13 @@ suite = {
       "sourceUrls" : ["https://lafo.ssw.uni-linz.ac.at/pub/jmh/jmh-runner-1.12-sources.jar"],
       "licence" : "GPLv2-CPE",
     },
+
+    "JACKPOT" : {
+      "urls" : [
+        "https://lafo.ssw.uni-linz.ac.at/pub/graal-external-deps/jackpot-8.1-20151011.220626.jar",
+      ],
+      "sha1" : "b5f91770afd3b8ce645e7b967a1f266ab472053b",
+    },
   },
 
   "licenses" : {
--- a/mx.py
+++ b/mx.py
@@ -68,6 +68,7 @@ import mx_unittest
 import mx_findbugs
 import mx_sigtest
 import mx_gate
+import mx_jackpot
 import mx_compat
 import mx_microbench
 import mx_urlrewrites
@@ -13838,6 +13839,7 @@ _commands = {
     'ideclean': [ideclean, ''],
     'ideinit': [ideinit, ''],
     'intellijinit': [intellijinit, ''],
+    'jackpot': [mx_jackpot.jackpot, ''],
     'jacocoreport' : [mx_gate.jacocoreport, '[output directory]'],
     'archive': [_archive, '[options]'],
     'maven-install' : [maven_install, ''],
@@ -14222,7 +14224,7 @@ def main():
         # no need to show the stack trace when the user presses CTRL-C
         abort(1)
 
-version = VersionSpec("5.41.1")
+version = VersionSpec("5.42.0")
 
 currentUmask = None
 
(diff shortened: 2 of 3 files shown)


mx_jackpot.py
```diff
@@ -0,0 +1,101 @@
...
+import tempfile
+import zipfile
+import shutil
+from os.path import join, exists
+
+def _should_test_project(p):
+    if not p.isJavaProject():
+        return False
+    if hasattr(p, 'jackpot'):
+        return p.findbugs.lower() == 'true'
```
dougxc: I think this should be `return p.jackpot.lower() == 'true'`

mx.mx/suite.py
```diff
@@ -110,6 +110,14 @@
       "sourceUrls" : ["https://lafo.ssw.uni-linz.ac.at/pub/jmh/jmh-runner-1.12-sources.jar"],
       "licence" : "GPLv2-CPE",
     },
+
+    "JACKPOT" : {
+      "path" : "lib/jackpot-3.0-8.1-SNAPSHOT.jar",
```
dougxc: You should omit the `"path"` attribute so that mx can place the download in its cache (i.e., `~/.mx/cache`)

mx_jackpot.py
jlahoda: Ooops, yes, thanks!

mx.mx/suite.py
jlahoda: Thanks, done!

dougxc:
You can fix the OCA check by committing with your Oracle email address.

mx.py
```diff
@@ -13841,6 +13842,7 @@ def warn(msg, context=None):
     'ideclean': [ideclean, ''],
     'ideinit': [ideinit, ''],
     'intellijinit': [intellijinit, ''],
+    'jackpot': [mx_jackpot.jackpot, ''],
```
dougxc: You should also bump the minor version number in the global `version` field near the bottom of `mx.py`.

mx_jackpot.py
```diff
@@ -0,0 +1,99 @@
...
+        jackpotJar = join(jackpotHome, 'jackpot.jar')
+    else:
+        jackpotJar = mx.library('JACKPOT').get_path(resolve=True)
+    assert exists(jackpotJar)
+    if suite is None:
+        suite = mx._primary_suite
+    nonTestProjects = [p for p in mx.projects() if _should_test_project(p)]
+    if not nonTestProjects:
+        return 0
+    cache = tempfile.mkdtemp(prefix='jackpot-mx-cache', dir=mx._mx_suite.dir)
```
dougxc: This cache doesn't appear to be used?

jlahoda:
I've rebased to current upstream mx master, and reflected the above comments (thanks for them!).

dougxc:
Jan, is there an immediate use case for this capability in one of the projects using mx? I'd rather not add a dependency in mx to another server (i.e., http://lahoda.info) unless the associated feature is (or will soon be) used by more than one project.

jtulach:
We are already using the Jackpot features in the Truffle API to simplify upgrade from deprecated methods to their more modern replacements. See the `.hint` files at https://github.com/graalvm/truffle/tree/master/truffle/com.oracle.truffle.api/src/META-INF/upgrade and elsewhere.

With these `.hint` files the Jackpot can fix your code automatically. The only drawback right now is: you have to start NetBeans to do so.

I know that most of the developers are using Eclipse and thus I asked Jan to contribute this command line integration so it is available to everyone and we can even make it part of the gate.

jlahoda:
Regarding the dependency on the lahoda.info server - I was hoping that (eventually), the binary could be added to the same binary server as the other binaries.

(Thanks for all the comments, BTW!)

dougxc:
I will work with our university collaborators to host the binary at https://lafo.ssw.uni-linz.ac.at/pub/graal-external-deps/jackpot-8.1-20151011.220626.jar if you can please update the URL in `suite.py`.

PR #121, opened 2016-09-14
Fix UnicodeEncodeError when running "mx checkstyle" on german locale

pointhi:


mx.py
```diff
@@ -10114,7 +10114,7 @@ def start_element(name, attrs):
                             if name == 'file':
                                 source[0] = attrs['name']
                             elif name == 'error':
-                                errors.append('{0}:{1}: {2}'.format(source[0], attrs['line'], attrs['message']))
+                                errors.append('{0}:{1}: {2}'.format(source[0].encode('utf8'), attrs['line'], attrs['message'].encode('utf8')))
```
rschatz: This breaks on non-utf8 locales (or at least produces garbled output):
```
> LANG=de_AT.utf-8 xterm
> mx checkstyle
...
StubAVXTest.java:80: Überflüssiger Modifier 'public'.
...
```

```
> LANG=de_AT xterm
> mx checkstyle
...
StubAVXTest.java:80: ÃberflÃ¼ssiger Modifier 'public'.
...
```

Correct fix:
`errors.append(u'{0}:{1}: {2}'.format(source[0], attrs['line'], attrs['message']))`

This puts a "unicode string" instead of an "8-bit string" into the errors array, and lets the `print` function (inside `log`) deal with the encoding. That works on both the `de_AT` and the `de_AT.utf-8` locale.

rschatz:
Rebased: https://github.com/graalvm/mx/commit/f28cad0fa5fe989fc7fb4dca6e840a8cbdc99c02

PR #122, opened 2016-11-09
Use custom Eclipse workspace properly

tmysik:
I do not use Eclipse so I don't like that running ```mx eclipseformat``` creates the ```workspace``` directory in my home every time  :) This change should fix using custom workspace (```WORKSPACE``` environment variable).

Thanks.
PS: I do not know Python so the fix can be perhaps not the best one or even incorrect.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -9638,7 +9638,7 @@ def eclipseformat(args):
     if not os.access(args.eclipse_exe, os.X_OK):
         abort('Not an executable file: ' + args.eclipse_exe)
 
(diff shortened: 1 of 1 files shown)


dougxc:
Hi Tomas, thanks for the patch. Once you can confirm you have signed the OCA, we can integrate it.

tmysik:
Done by commit https://github.com/graalvm/mx/commit/899d0db089b075c639804af1389aa36cec98eb0b.

Thank you.

PR #124, opened 2016-12-21
Added a feature for native projects and fixed a bug for tar distributions

blu3r4y:
Those two commits are independent of each other.

One enables to pass the target argument to gnu make and the other fixes an issue where tar distributions were not removed by mx clean.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -1494,6 +1494,9 @@ class TARArchiveTask(ArchiveTask):
             abort('should not reach here')
         if exists(self.subject.path):
             os.remove(self.subject.path)
+        tarOutputDir = join(self.subject.suite.dir, self.subject.output)
+        if os.path.exists(tarOutputDir):
+            shutil.rmtree(tarOutputDir)
 
     def cleanForbidden(self):
         if ArchiveTask.cleanForbidden(self):
--- a/mx.py
+++ b/mx.py
@@ -1415,7 +1415,7 @@ class NativeTARDistribution(Distribution):
         filesToArchive = []
         for d in self.archived_deps():
             if d.isNativeProject():
-                filesToArchive.extend(self.get_files_for_nativeProject(d, False))
+                filesToArchive.extend(self.get_files_for_nativeProject(d, reportOutput))
             elif d.isArchivableProject():
                 filesToArchive.extend(self.get_files_for_archivableProject(d))
             elif hasattr(d, 'getResults') and not d.getResults():
--- a/mx.py
+++ b/mx.py
@@ -1518,11 +1518,12 @@ class TARArchiveTask(ArchiveTask):
         if exists(self.subject.path):
             os.remove(self.subject.path)
 
-        if self.subject.output:
-            for _, arcname in self.subject.get_files_to_archive(False):
-                dest = join(self.subject.suite.dir, self.subject.output, arcname)
-                if exists(dest):
-                    os.remove(dest)
+        if not forBuild:
+            if self.subject.output:
+                for _, arcname in self.subject.get_files_to_archive(False):
+                    dest = join(self.subject.suite.dir, self.subject.output, arcname)
+                    if exists(dest):
+                        os.remove(dest)
 
     def cleanForbidden(self):
         if ArchiveTask.cleanForbidden(self):
--- a/mx.py
+++ b/mx.py
@@ -2810,6 +2810,8 @@ class NativeBuildTask(ProjectBuildTask):
             cmdline += ['-f', join(self.subject.dir, 'Makefile')]
         else:
             cwd = self.subject.dir
+        if hasattr(self.subject, "makeTarget"):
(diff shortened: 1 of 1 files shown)


blu3r4y:
I restructured the code, such that a method within `NativeTARDistribution` exists, which returns all the files that will be copied by `archive_and_copy`. `clean` will use this method to only remove files that where copied.

mx.py
```diff
@@ -1406,42 +1406,65 @@ def archive_and_copy(name, arcname):
...
+            for name, arcname in self.get_files_to_archive():
+                archive_and_copy(name, arcname)
 
         self.notify_updated()
 
+    def get_files_to_archive(self, reportOutput=True):
+        filesToArchive = []
+        for d in self.archived_deps():
+            if d.isNativeProject():
+                filesToArchive.extend(self.get_files_for_nativeProject(d, False))
```
rschatz: Shouldn't this be `get_files_for_nativeProject(d, reportOutput)`?

blu3r4y: Yes, it should. Thank you for that!

blu3r4y:
The previous build failed, because of a dead link (see #126). I rebased the pull request, should be fine now.

rschatz:
There is still a weird failure if you set the output directory to `.`. I'm attaching a repro case:
[mx-test.tar.gz](https://github.com/graalvm/mx/files/744936/mx-test.tar.gz)

This dummy project builds cleanly on the revision that this branch is based on, but fails on this PR:
```
Archiving TEST_NATIVE... [dependency test.native updated]
Process Process-4:
  File "/usr/lib/python2.7/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/usr/lib/python2.7/multiprocessing/process.py", line 114, in run
    self._target(*self._args, **self._kwargs)
  File "/home/roland/git/mx/mx.py", line 9868, in executeTask
    task.execute()
  File "/home/roland/git/mx/mx.py", line 708, in execute
    self.build()
  File "/home/roland/git/mx/mx.py", line 1329, in build
    self.subject.make_archive()
  File "/home/roland/git/mx/mx.py", line 1410, in make_archive
    archive_and_copy(name, arcname)
  File "/home/roland/git/mx/mx.py", line 1401, in archive_and_copy
    arc.zf.add(name, arcname=arcname)
  File "/usr/lib/python2.7/tarfile.py", line 2009, in add
    tarinfo = self.gettarinfo(name, arcname)
  File "/usr/lib/python2.7/tarfile.py", line 1881, in gettarinfo
    statres = os.lstat(name)
OSError: [Errno 2] No such file or directory: '/home/roland/test/mx-test/./test-output-file'
Archiving TEST_NATIVE failed
1 build tasks failed
```

blu3r4y:
The issue arises because the clean task gets called during some builds, which cleans files needed by the archive task later on. The clean happens [on this line within mx.py](https://github.com/graalvm/mx/blob/ecc49e995882fc0813f91944ae9f952b847be9e5/mx.py#L706).

I assume the behavior of a "clean-for-build-call" is needed for some projects in some way. Thus, I changed the clean method for native projects, such that it will not clean any files, if the clean method is called with `forBuild = true`.

gilles-duboscq:
> The issue arises because the clean task gets called during some builds, which cleans files needed by the archive task later on. The clean happens on this line within mx.py.

That seems to go against the overall design: a built task should have separate inputs and outputs. It should always be possible to delete those outputs (i.e., clean) before (re-)building. A failure there indicates that output and inputs are somehow mixed.

blu3r4y:
Well, that seems to me like a workaround in `BuildTask` is broken now because of a correct `clean`.

However, to fix this, i just have to take care of the `forBuild` argument in the `clean(forBuild)` call. By the way, this is also done in the `NativeBuildTask` [here](https://github.com/graalvm/mx/blob/ecc49e995882fc0813f91944ae9f952b847be9e5/mx.py#L2851-L2859).

My latest commit 4505fe1866338e3d8c947675fb54403e8d1f981f does exactly that. So it won't break a `clean` within `NativeBuildTask` and one can still perform a sound clean with an explicit `mx clean`.

blu3r4y:
Any updates on that? Can we integrate it with fix 4505fe1866338e3d8c947675fb54403e8d1f981f?

gilles-duboscq:
I still think the required workaround points at a problem (which might be an existing problem, not necessarily a problem in your changes) and i would like to look into it before changing more things around native projects. Unfortunately i didn't have much time to look at it. The reason i'd prefer to be sure before merging this is that once we merge this we have to remain backwards compatible with the current behaviour.

I'll try to make some time to look into it this week.

blu3r4y:
I see your point. If the change is too problematic, don't struggle with other issues arising from it. This feature is not mission critical to me ;)

blu3r4y:
I squashed my commits and made a new PR #129.

PR #125, opened 2017-01-02
Write colored log output (at least on UNIX systems)

pointhi:
Currently, spotting the error line in bigger amounts of terminal output is quite a pain.

This commit adds the new function ```log_error```, which handles such message in an improved way:
* write message to stderr instead of stdout
* print the string in red color on the terminal (at least for unix terminals now)

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -9364,6 +9364,21 @@ def log(msg=None):
     else:
         print msg
 
+def log_error(msg=None):
+    """
+    Write an error message to the console.
+    All script output goes through this method thus allowing a subclass
+    to redirect it.
+    """
+    isUnix = sys.platform.startswith('linux') or sys.platform in ['darwin', 'freebsd']
+    if msg is None:
+        sys.stderr.write(os.linesep)
+    elif isUnix and sys.stderr.isatty():
+        # On unix systems, we can use ANSI escape sequences
+        sys.stderr.write('\033[91m' + msg.encode('utf8') + '\033[0m' + os.linesep)
+    else:
+        sys.stderr.write(msg.encode('utf8') + os.linesep)
+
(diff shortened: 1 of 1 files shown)


mx.py
```diff
@@ -9341,6 +9341,21 @@ def log(msg=None):
...
 
+def log_error(msg=None):
+    """
+    Write an error message to the console.
+    All script output goes through this method thus allowing a subclass
+    to redirect it.
+    """
+    if msg is None:
+        sys.stderr.write(os.linesep)
+    elif sys.platform.startswith('linux') or\
```
dougxc: no need to split such short lines

mx.py
```diff
@@ -9341,6 +9341,21 @@ def log(msg=None):
...
+         sys.platform in ['darwin', 'freebsd']:
+        # On unix systems, we can use ANSI escape sequences
+        sys.stderr.write('\033[91m' + msg.encode('utf8') + '\033[0m' + os.linesep)
```
dougxc: How does this look in captured log files? Maybe we should only colorize if in an interactive console.

pointhi:
@dougxc thanks for your input.

I changed the PR accordingly

dougxc:
I assume the encoding being performed is to mimic what the `print` statement does. In that case, let's just use `print` directly:
```
    if msg is None:
        print >> sys.stderr
    elif isUnix and sys.stderr.isatty():
        # On unix systems, we can use ANSI escape sequences
        print >> sys.stderr, '\033[91m' + msg + '\033[0m'
    else:
        print >> sys.stderr, msg
```
If you're ok with that, I'll integrate it.

pointhi:
No problem, you can integrate it.

PR #126, opened 2017-01-10
Use ecj-4.5.2.jar from eclipse archive for travis

blu3r4y:
The original resource at http://ftp.halifax.rwth-aachen.de/eclipse/eclipse/downloads/drops4/R-4.5.2-201602121500/ecj-4.5.2.jar is not available anymore and travis builds fail because of that. I suggest to use https://lafo.ssw.uni-linz.ac.at/pub/sulong-deps/ecj-4.5.2.jar instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,5 +15,5 @@ install:
       tar -C $TRAVIS_BUILD_DIR/.. -xf $ECLIPSE_TAR
       export ECLIPSE_EXE=$TRAVIS_BUILD_DIR/../eclipse/eclipse
       export JDT=$TRAVIS_BUILD_DIR/ecj.jar
-      wget http://ftp.halifax.rwth-aachen.de/eclipse/eclipse/downloads/drops4/R-4.5.2-201602121500/ecj-4.5.2.jar -O $JDT
+      wget http://www.eclipse.org/downloads/download.php?file=/eclipse/downloads/drops4/R-4.5.2-201602121500/ecj-4.5.2.jar -O $JDT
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -15,5 +15,5 @@ install:
       tar -C $TRAVIS_BUILD_DIR/.. -xf $ECLIPSE_TAR
       export ECLIPSE_EXE=$TRAVIS_BUILD_DIR/../eclipse/eclipse
       export JDT=$TRAVIS_BUILD_DIR/ecj.jar
-      wget http://ftp.halifax.rwth-aachen.de/eclipse/eclipse/downloads/drops4/R-4.5.2-201602121500/ecj-4.5.2.jar -O $JDT
+      wget https://lafo.ssw.uni-linz.ac.at/pub/sulong-deps/ecj-4.5.2.jar -O $JDT
```
dougxc: Is there no other official, reliable Eclipse mirror we can pull from?

blu3r4y: We could use http://www.eclipse.org/downloads/download.php?file=/eclipse/downloads/drops4/R-4.5.2-201602121500/ecj-4.5.2.jar instead. Would that be more reliable?

blu3r4y: I think this archive link would be best: http://archive.eclipse.org/eclipse/downloads/drops4/R-4.5.2-201602121500/download.php?dropFile=ecj-4.5.2.jar

dougxc: Maybe - we'll try it for now.

PR #128, opened 2017-02-16
Fix error when trying to print unicode strings

pointhi:
Sadly, python has quite a bunch of handling issues regarding utf-8

I solved the error in this specific way, because it's possible log_error is called with some object where we want to have the string representation, and not directly using str or unicode.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -9530,6 +9530,8 @@ def log_error(msg=None):
     if msg is None:
         print >> sys.stderr
     else:
+        if type(msg) is unicode:
+            msg = msg.encode('utf-8')
         print >> sys.stderr, colorize(str(msg), stream=sys.stderr)
 
 def expand_project_in_class_path_arg(cpArg):


dougxc:
What error did you hit that this fix addresses?

pointhi:
Well, It required some time before the error reoccur:

```
Checkstyle ends with 8 errors.
  File "/home/thomas/JKU/sulong-dev/mx/mx.py", line 15048, in <module>
    main()
  File "/home/thomas/JKU/sulong-dev/mx/mx.py", line 15032, in main
    retcode = c(command_args)
  File "/home/thomas/JKU/java-llvm-ir-builder-dev/sulong/mx.sulong/mx_sulong.py", line 373, in runChecks
    optionalRetValue = command(vmArgs)
  File "/home/thomas/JKU/sulong-dev/mx/mx.py", line 10802, in checkstyle
    map(log_error, errors)
  File "/home/thomas/JKU/sulong-dev/mx/mx.py", line 9533, in log_error
    print >> sys.stderr, colorize(str(msg), stream=sys.stderr)
UnicodeEncodeError: 'ascii' codec can't encode character u'\xf6' in position 185: ordinal not in range(128)
```

PR #130, opened 2017-03-08
Add option to specify make target explicitly

blu3r4y:


dougxc:
Can you please motivate this change? I would have thought that a native project being managed by mx can always arrange to have the default make target be the right one.

blu3r4y:
Of course, we encounter the following situation with some native project in Espresso:

The automatic build will use a environment variable, which specifies the target file directly. However, if I would like to test something locally in this folder, it would be convenient, to not specify this environment variable and just use a default name for the target.

```
mx: ${TARGET}
  # ...

all: bin/defaultname.so
  # ...
```

dougxc:
What do you mean by "in this folder"? What environment variable does the automatic build use? Sorry, I'm not that familiar with the native project support in mx so don't quite understand the problem you're trying to address.

blu3r4y:
We are building native libraries for Espresso, which are used for JNI calls in a Java project. So the library, which is built by `make` needs to be referenced properly (`java.library.path` variable), when executing the Java project.

In order to achieve this, we set an environment variable, which is passed on to the `make` process and specifies the desired libraries name. ([using `buildEnv`](https://github.com/graalvm/mx/blob/8591f10088f9020aee4b854a766b05a8872924f7/mx.py#L2827)) This library can then be referenced for the Java project. ([using `javaProperties`](https://github.com/graalvm/mx/blob/8591f10088f9020aee4b854a766b05a8872924f7/mx.py#L567))

That works perfectly fine for building and testing with `mx` alone. The Makefile has to use the environment variable as its output target for this to work.

The problem is, I occasionally would like to work directly with the native projects folder, and execute `make` from the command line manually during development. I don't want to trigger a huge `mx build` for that. It would just be a matter of convenience for me, to type `make` without specifying a custom target or setting the needed environment variable myself.

dougxc:
If you're typing `make` directly on the command line, why is an mx change needed at all? I'm not against this change but would still like to understand better.

blu3r4y:
I want to be able to use `mx build` (on whole suite) and `make` (for some project) from the console.

`mx build` will pass the desired output filename via an environment variable to the `make` process. Thereby, I can conveniently specify the output filename within the `suite.py` file.

However, I also want to be able to quickly build the native project with a simple `make` call from the command line, **without** setting the environment variable (just use some default name). But, because `mx` will always choose the default target, that needs to be the one with the environment variable in it - which is inconvenient for simple command line usage.

dougxc:
Thank you - now I understand ;-)

rschatz:
mx version bumped to `5.75.2`, please use at least this version in your `suite.py` if you want to use this feature.

PR #133, opened 2017-04-26
Add JUnit config to intellijinit

pejovica:
Makes intellijinit generate default run/debug configuration for JUnit.

There are two aspects to proposed changes. First one is the placement of generated configuration. At the moment it is stored in the workspace.xml, since the official location is not available yet (IDEA-65915). And the second one is the collection of necessary VM parameters for JUnit to run. I tried to reuse mx unittest for that by effectively making it execute a dry run. This implementation of dry run only prevents tests from running. It doesn't prevent side effects, e.g., unittest's test discovery and caching. But since unittest allows custom launchers, i.e., implemented by other modules, I am not sure what can be expected from them in this regard. Graal core seems to work fine.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -13172,20 +13172,18 @@ def _intellij_suite(args, s, refreshOnly=False, mx_python_modules=False, java_mo
 
                 class DisabledUnittests(object):
                     """Context manager that temporarily disables unittests from launching.
-                    
+
                     It does this by replacing the mx.run with itself and examining the issued
                     commands. If it detects the `mx_unittest_main_class` in the command, it
                     captures the command and reports success without actually doing anything.
                     """
 
                     def __enter__(self):
-                        global run
                         self.run = run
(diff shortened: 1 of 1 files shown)


vjovanov:
@pejovica could you please sign the oracle contributor agreement? Here are the instructions: 
  http://www.oracle.com/technetwork/community/oca-486395.html
Also, the gate is failing due to style issues. To reproduce locally run: 
```
mx --strict-compliance gate --strict-mode
```

vjovanov:
Otherwise, LGTM

pejovica:
I've added the final set of changes, so could you have another look?

vjovanov (changes requested):

vjovanov:
@pejovica will there be progress on this or we can close the PR?

vjovanov:
@pejovica is this still valid?

PR #135, opened 2017-05-01
Fix maven-install

okutane:
Not all projects have 'maven' attribute, this results in error during maven-install command invocation (can be reproduced on 'sulong' project)


okutane:
not needed? :)

dougxc:
@grimmerm can you please sponsor this after making sure it solves the sulong project issue mentioned in the description.

PR #137, opened 2017-08-23
Make log() "thread-safe" to avoid empty/joined lines

### LINKED ISSUE #136: Empty lines on a multi-core system
On a multi-core system (with Python 2.7.13), mx occasionally prints empty lines:
```
…
Building com.oracle.truffle.nfi.native with GNU Make... [dependency com.oracle.truffle.nfi updated]
Archiving TRUFFLE_DEBUG... [dependency com.oracle.truffle.tools.debug.shell updated]
Compiling com.oracle.truffle.tck with ecj-daemon... [dependency TRUFFLE_DSL_PROCESSOR updated]
Compiling com.oracle.truffle.api.interop.java.test with ecj-daemon... [dependency TRUFFLE_DSL_PROCESSOR updated]JAVAH src_gen/native.h

Compiling com.oracle.truffle.api.dsl.test with ecj-daemon... [dependency TRUFFLE_DSL_PROCESSOR updated]
CC bin/closure.o
CC bin/jni.o
CC bin/lookup.o
CC bin/signature.o
Archiving TRUFFLE_TCK... [dependency com.oracle.truffle.tck updated]
…
```

pekd:
This should fix #136.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -9856,7 +9856,7 @@ def log(msg=None):
     if msg is None:
         print
     else:
-        print msg
+        print msg + "\n",
 
 # https://en.wikipedia.org/wiki/ANSI_escape_code#Colors
 _ansi_color_table = {
--- a/mx.py
+++ b/mx.py
@@ -9856,6 +9856,19 @@ def log(msg=None):
     if msg is None:
         print
     else:
+        # https://docs.python.org/2/reference/simple_stmts.html#the-print-statement
+        # > A '\n' character is written at the end, unless the print statement
+        # > ends with a comma.
+        #
+        # In CPython, the normal print statement (without comma) is compiled to
+        # two bytecode instructions: PRINT_ITEM, followed by PRINT_NEWLINE.
+        # Each of these bytecode instructions is executed atomically, but the
+        # interpreter can suspend the thread between the two instructions.
+        #
+        # If the print statement is followed by a comma, the PRINT_NEWLINE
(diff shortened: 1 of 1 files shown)


mx.py
```diff
@@ -9856,7 +9856,7 @@ def log(msg=None):
     if msg is None:
         print
     else:
-        print msg
```
dougxc: Won't this just cause a lot of empty lines? Or does the `print` statement in python not add a `\n` if there's already one in `msg`?

pekd: The short version: no, because magic happens because of the `,`.

The long version: [the documentation](https://docs.python.org/2/reference/simple_stmts.html#the-print-statement) states that
> A `'\n'` character is written at the end, unless the print statement ends with a comma.

In CPython, the normal print statement (without comma) is compiled to two bytecode instructions: `PRINT_ITEM`, followed by `PRINT_NEWLINE`. Each of these bytecode instructions is executed atomically, but the interpreter can suspend the thread between the two instructions.

If the print statement is followed by a comma, the `PRINT_NEWLINE` instruction is omitted. By manually adding the newline to the string, there is only a single `PRINT_ITEM` instruction which is executed atomically, but still prints the newline.

woess: oh wow, that's really unfortunate. I guess the print *function* does not suffer from this problem?

dougxc: That kind of python magic definitely deserves a comment. Can you please add your above description as a comment.

PR #140, opened 2018-01-07
Fix IntelliJ order entries order

afdw:
Without this change IDEA tries to use classes from JDK with same names, which results in resolve errors.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -13638,7 +13638,6 @@ def _intellij_suite(args, s, declared_modules, referenced_modules, refreshOnly=F
                     _intellij_exclude_if_exists(moduleXml, p, name, output=True)
                 moduleXml.close('content')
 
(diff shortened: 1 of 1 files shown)


vjovanov:
Hi @afdw, thanks for your contribution.

In order to merge this PR you will first need to sign the Oracle OCA: 

  http://www.oracle.com/technetwork/community/oca-486395.html

Once you do that the `oca-check` will pass.

In the meanwhile, I will take care of the style issue in `travis` which is unrelated to this PR.

vjovanov:
We have fixed the style issue. After rebasing the CI job should pass.

afdw:
@vjovanov thanks, yes.

vjovanov:
All good, merging the code.

afdw:
What is `GR-xxxx`? Is it private bug tracker? Or where is it located?

vjovanov:
Yes, we make internal issues for every PR and hence every external issue. The internal issue has no additional information: just contains a link to this issue.

PR #142, opened 2018-01-14
Generate native headers during java compilation

dsamersoff:
Javah deprecated since jdk8 and removed in jdk11, so we have to replace it to javac -h <header_output_directory> in a build and move header generation from native build phase to java build phase.

With proposed changes mx will always generate headers if Java class contains native method.


rschatz:
Included in #143.

PR #144, opened 2018-02-12
Allow newer versions of pylint

fniephaus:
I don't know why the version of pylint is locked to `1.1.x`, but I'm using the latest pylint `1.8.2` without any problems. This change allows newer versions of pylint as well.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -11199,8 +11199,8 @@ def pylint(args):
             log_error('could not determine pylint version from ' + output)
             return -1
         major, minor, micro = (int(m.group(1)), int(m.group(2)), int(m.group(3)))
-        if major != 1 or minor != 1:
-            log_error('require pylint version = 1.1.x (got {0}.{1}.{2})'.format(major, minor, micro))
+        if major != 1 or minor < 1:
+            log_error('require pylint version >= 1.1.x (got {0}.{1}.{2})'.format(major, minor, micro))
             return -1
     except BaseException as e:
         log_error('pylint is not available: ' + str(e))


fniephaus:
I've signed the OCA and my name is on the list. I think the problem was that I created the commit on GitHub...I've recreated the commit locally with the correct committer email and force pushed.

gilles-duboscq:
The reason the version is locked is to avoid diagnostics introduced after 1.1.x to trigger. If there is a way to do that with newer version we could support that but last time i looked it was not really possible without manually compiling a lists of things to disable for each new version.

zapster:
@fniephaus do you actually *want* features from `>1.1.x` or is it just out of convenience because the system default is `1.8.x`? If the former, we might want to introduce support for specifying a checkstyle version in the `suite.py` for selected projects. However this might require some work.
If not then setting up a `virtualenv` with pylint `1.1.0` is IMHO a simpler workaround.

fniephaus:
@gilles-duboscq thanks, that makes sense then.

@zapster it's more out of convenience really, didn't know about the reason for the version lock. Will use a `virtualenv` instead.

gilles-duboscq:
I really wish they had annotated each rule with the version in which it was introduced... Otherwise getting reproducible results across versions is hard.

In the end, to get reproducible results in our CI we have to first install `astroid==1.1.0` and *then* install `pylint==1.1.0`. Otherwise you might also get a newer version of `astroid` which causes some other problems.

fniephaus:
Thanks for the info, @gilles-duboscq.

PR #145, opened 2018-02-12
Add ability to export Jacoco reports as XML

fniephaus:
This allows to export Jacoco reports as XML. This way, it is possible to use coverage services like coveralls or codecov which don't seem to support the `jacoco.exec` format, but its XML format.
The new code does not change the default behavior (HTML output).

Example: `mx jacocoreport --format xml ./` generates an XML in the current directory.


fniephaus:
Any update on this?

dougxc:
Sorry for the delay. This change looks good. I'll put it through our CI and integrate it.

fniephaus:
Thanks, @dougxc!

PR #156, opened 2018-07-21
Feature Request: support compilation from tarball [PoC, don't merge]

smarr:
This is more of a feature request than a PR.

I'd like to be able to build Graal from a source tarball for the use in [artifacts](http://www.artifact-eval.org/).
The attached patch introduces a stub for a VC in mx, that does achieve that.

However, it's not a general solution, because the suite structure is somewhat hard coded.

Nonetheless, it supports the following use case:

```bash
cd graal-jvmci-8
mx clean
mx build
export JAVA_HOME=`../mx/mx jdkhome`
cd ..

cd graal/compiler
export JVMCI_VERSION_CHECK=ignore
mx clean
mx build
mx makegraaljdk $OUT/graal-core
```

If this is something that would be generally useful, any idea how this could be realized less hacky?

Thanks!


gilles-duboscq:
Something like that would be useful.
In principle `mx` should support the case where the `vc` and `vc_dir` fields of the `Suite` are `None`.
I'm not sure what's easier though: make an artificial `VC` or fix the code that accesses `vc` and `vc_dir`.

As you note in your commit, there needs to be some meta-data in the `suite.py` file to figure out at what depth the root of the "repo" is.

PR #134, opened 2017-04-26
Fix intellijinit on Windows

okutane:
Hi! I 've just tried to run 'mx.cmd intellijinit' for sulong project on windows and got following error:
```Traceback (most recent call last):
  File "T:\dev\sulong-dev\mx\/mx.py", line 15330, in <module>
    main()
  File "T:\dev\sulong-dev\mx\/mx.py", line 15314, in main
    retcode = c(command_args)
  File "T:\dev\sulong-dev\mx\/mx.py", line 12737, in intellijinit
    _intellij_suite(args, suite, refreshOnly, mx_python_modules, java_modules, suite != primary_suite())
  File "T:\dev\sulong-dev\mx\/mx.py", line 12959, in _intellij_suite
    path = os.path.relpath(library.get_path(True), s.dir)
  File "C:\Python27\lib\ntpath.py", line 529, in relpath
    % (path_prefix, start_prefix))
ValueError: path is on drive C:, start on drive T:```

The problem occured because I've put my dev directory on T: drive and temp files were under my user directory on C:

I've implemented the minimal fix for my scenario which shouldn't break anything and afterwards reviewed neighboorhood code and added few similar fixes "just in case".

This is my first PR to mx and first PR in Python, so please give any necessary feedback on how to improve it and get it merged. :)

PR #164, opened 2018-11-08
Fix os.path.relpath throwing

JornVernee:
Replaced calls to os.path.relpath with a safe alternative, since not every path can be made relative on Windows, due to different drives.

This is based on another PR [1], where the author asked someone else to take over since they don't have time to fix the merge conflicts [2].

~I tried running `mx gate`  but that throws an unrelated error, so it doesn't look like that works on Windows yet. So be aware that this has not passed gate yet.~

[1] : https://github.com/graalvm/mx/pull/134
[2] : https://github.com/graalvm/mx/pull/134#issuecomment-436805868

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -15768,7 +15768,7 @@ def _intellij_suite(args, s, declared_modules, referenced_modules, sdks, refresh
         if not module_files_only:
             declared_modules.add(basename(s.mxDir))
             moduleFilePath = "$PROJECT_DIR$/" + os.path.relpath(moduleFile, s.dir)
-            modulesXml.element('module', attributes={'fileurl': 'file://$' + moduleFilePath, 'filepath': moduleFilePath})
+            modulesXml.element('module', attributes={'fileurl': 'file://' + moduleFilePath, 'filepath': moduleFilePath})
 
             declared_modules.add(basename(_mx_suite.dir))
             mxModuleFile = join(_mx_suite.dir, basename(_mx_suite.dir) + '.iml')
--- a/mx.py
+++ b/mx.py
@@ -227,6 +227,17 @@ class DynamicVarScope(object):
 
 currently_loading_suite = DynamicVar(None)
 
+def relpath_or_absolute(path, start, prefix=""):
+    """
+    Finds a relative path and joins it to 'prefix', or otherwise tries to use 'path' as an absolute path.
+    If 'path' is not an absolute path, an error is thrown.
+    """
+    try:
+        return join(prefix, os.path.relpath(path, start))
+    except ValueError:
+        if not os.path.isabs(path):
+            raise ValueError('can not find a relative path to dependency and path is not absolute: ' + path)
+        return path
 
 # Support for Python 2.6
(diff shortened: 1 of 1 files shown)


JornVernee:
I have already signed the OCA: https://openjdk.java.net/census#jvernee

JornVernee:
Hey Doug,

Just to make sure I'd like to wait for the old author to give the OK before merging this.

It seems like they have abandoned their own PR, and don't really care, but they _did_ come up with the fix after all.

dougxc:
I think we're fine to proceed based on @okutane 's [comment](
https://github.com/graalvm/mx/pull/134#issuecomment-436805868).

JornVernee:
@dougxc Ok, then go ahead.

gilles-duboscq:
Hi @JornVernee, thanks for taking a stab at this.
Unfortunately we can not simply replace `os.path.relpath` with `maybe_relpath`. I would actually say that we can only do it in rare cases.

In particular, there are cases where the logic simply doesn't work if a relative path can not be created, in those it needs to remain an error and the root case has to be fixed.

In many cases it's OK anyway because relapth is used in conditions where a relative path can always be found (in particular when using `os.walk`).

In other cases, a more complex logic is needed. For example in intellij project generation, when a relative path can be found, the `$PROJECT_DIR$` variable should be used and if it can not then an absolute path should be used.

In other places a unix path is used as a short-cut to something else (a relative URL, a logical path in an archive etc.) and an other platform-independent approach should be used.

In order to make progress, I propose that we don't fix everything at once but just fix the ones where `maybe_relpath` can be used and the `intellijinit` ones since that was the purpose of the original PR.

JornVernee:
I see travis is running `mx gate` in strict mode.

When doing that myself I get:
```
require pylint version = 1.1.x (got 1.8.4)
Pylint not configured correctly. Cannot execute Pylint task.
```
Should that requirement be bumped up to 1.x.x ?

JornVernee:
I have successfully run `mx ideinit` on the graal/compiler suite. I also tried the vm suite, but that failed due to an unrelated error.

JornVernee:
I think I/we will have to look into this some more. One other problem I realized is that the paths are sometimes used as URLs, for which they must use `/` as a path separator. On unix that just works out when using normal paths, but on Windows it does not.

JornVernee:
I looked into this and decided to keep the changes simple, @gilles-duboscq or @dougxc  can you take another look? I pushed the relpath lookup into make_library so it can either use `$PROJECT_DIR$` or an absolute path. (I squashed all my commits, so the commit message looks a bit strange)

I was able to pass `mx build` on the graal/compiler suite and run it with `mx vm` so that's 👍 . I also ended up fixing #165 in the process (a regex that wasn't matching on Windows path separators).

gilles-duboscq:
> Any idea where I can get the right version of pylint?
```
$ pip install 'astroid==1.1.0'
$ pip install 'pylint==1.1.0'
```
(You have to install strict versions and install astroid first with a strict version although it's an explicit dependency of pylint)

PR #167, opened 2018-11-08
Add missing has_documentation to MxCommand

JornVernee:
Running `mx help <command>` throws `'MxCommand' object has no attribute 'has_documentation'`. ~This PR adds that missing function.~

This PR:
- Replaces `has_documentation` with `get_doc`, since this simplifies usage.
- Makes sure the `unittest` command passes a doc function instead of a doc string (and removed extra newlin in doc string)
- Makes sure the `get_doc` function actually returns the documentation string.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -13679,16 +13679,7 @@ Given a command name, print help for that command."""
             abort('mx: command \'{0}\' is ambiguous\n    {1}'.format(name, ' '.join(hits)))
 
     command = _mx_commands.commands()[name]
-    doc = command.command_function.__doc__
-    if command.has_documentation():
-        format_args = []
-        if command.usage_msg:
-            format_args.append(command.usage_msg)
-        if command.doc_function:
-            format_args.append(command.doc_function())
-
-        doc = doc.format(*format_args)
-    print 'mx {0} {1}\n\n{2}\n'.format(name, command.usage_msg, doc)
+    print command.get_doc()
 
 def _parse_multireleasejar_version(value):
     try:
--- a/mx_commands.py
+++ b/mx_commands.py
@@ -125,6 +125,20 @@ class MxCommand(object):
     def command_function(self):
         return self._command_function
 
+    def get_doc(self):
+        doc = 'mx {0} {1}'
+        msg = '<no documentation>'
+        if self.command_function.__doc__ or self.doc_function or self.usage_msg:
+            msg = ''
+            if self.usage_msg:
+                msg += self.usage_msg
+            if self.command_function.__doc__:
+                msg += '\n\n' + self.command_function.__doc__
+            if self.doc_function:
+                msg += '\n' + self.doc_function()
+
+        return doc.format(self.command, msg)
+
     def __call__(self, *args, **kwargs):
         for callback in self._mx_commands.command_before_callbacks:
             callback(self, *args, **kwargs)
--- a/mx_unittest.py
+++ b/mx_unittest.py
@@ -312,7 +312,6 @@ def _unittest(args, annotations, junit_args, prefixCp="", blacklist=None, whitel
             os.remove(testfile)
 
 unittestHelpSuffix = """
-
     To avoid conflicts with VM options '--' can be used as delimiter.
 
     If test filters are supplied, only tests whose fully qualified name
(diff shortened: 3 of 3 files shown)


mx_commands.py
```diff
@@ -125,6 +125,9 @@ def __init__(self, mx_commands, command_function, suite_name, command, usage_msg
     def command_function(self):
         return self._command_function
 
+    def has_documentation(self):
+        return self.doc_function != None
```
gilles-duboscq: From the usage of `has_documentation`, it looks like this should also take the presence of a `usage_msg` into account.

JornVernee: Ah, you're right. Actually, from the usage I think maybe the `has_documentation` function could be replaced by a `get_doc` function that creates the doc that is being used there:

```python
# mx.py
def help_(args):
    # ...
    print 'mx {0} {1}\n\n{2}\n'.format(name, command.usage_msg, command.get_doc())
```
Where:
```python
# MxCommand
def get_doc(self):
    doc = self.command_function.__doc__
    format_args = []
    if self.usage_msg:
        format_args.append(self.usage_msg)
    if self.doc_function:
        format_args.append(self.doc_function())
    return doc.format(*format_args)
```

What do you think?

JornVernee:
Actually, looking at some of the use cases, like unittest https://github.com/graalvm/mx/blob/master/mx_unittest.py#L351 the documentation never makes it into the doc message because the function right now expects the command's doc string to be a format string.

I think I have a way to fix this.

dougxc:
Hi @JornVernee , can you please ensure all your commits are made with your real email address. This will avoid the complaints by the graalvmbot.

JornVernee:
@dougxc Ah ok. I purposely stopped adding my emails to the commits since I was getting a lot of recruitment spam.

I will add the email with the next commit.

dougxc:
Thanks and sorry to hear about the spam.

You'll also need to redo the commit (rebase) that is using the github email address.

mx.py
```diff
@@ -13654,16 +13654,7 @@ def help_(args):
...
-    if command.has_documentation():
-        format_args = []
-        if command.usage_msg:
-            format_args.append(command.usage_msg)
-        if command.doc_function:
-            format_args.append(command.doc_function())
-
-        doc = doc.format(*format_args)
-    print 'mx {0} {1}\n\n{2}\n'.format(name, command.usage_msg, doc)
+    print '\nmx {0} {1}\n\n{2}\n{3}\n'.format(name, command.usage_msg, command.command_function.__doc__, command.get_doc())
```
gilles-duboscq: If there's no documentation i think it would be better if it just prints the usage line.
Maybe only if there's nothing (no usage, no docstring, no doc function) should it print that there is no documentation.

I'm also not sure we need the leading `\n`

gilles-duboscq:
> Not sure why Travis is choking, I can not reproduce that error on my machine. Is something going on with the CI setup?

Seems to be an issue with the JDT (eclipse compiler/ECJ) JAR which comes from there:
https://github.com/graalvm/mx/blob/9c00c2ea9809bdf4fb24e26aec2c7aec8967e3b4/.travis.yml#L14

At least locally this URL works fine. I'm not sure what's going on. Maybe the download got corrupted?

JornVernee:
@gilles-duboscq I checked out the URL and there seems to be no re-direct to `https://` either (I had that problem a few days ago on another project)

mx.py
JornVernee: @gilles-duboscq I thought the leading `\n` looked nicer, since the help message wouldn't be squeezed up against the terminal prompt (if that makes sense).

JornVernee:
@gilles-duboscq Well, I implemented your changes and the travis check is passing as well now. My other PR had the same problem: https://github.com/graalvm/mx/pull/172 Is there any way you could re-run the check?

Thanks

gilles-duboscq:
Looks good now, thank you!

Regarding #172, maybe once this is integrated, you can rebase it? For some reason it seems to also contain the changes from #164 which has already been integrated so maybe a rebase is anyway a good idea.

gilles-duboscq:
Also this is changing a line with print so it will definitely conflict with #172.

JornVernee:
OK, I'll do that after this is integrated then. Then I can make sure there's no other conflicts or weird diffs with mx/master as well.

PR #171, opened 2018-11-09
Fix long paths in Windows.

jphalimi-intel:
This patch fixes issues in Windows for long-path related problems.


mx.py
```diff
@@ -3625,14 +3625,14 @@ def build(self):
                 for fname in filenames:
                     output.append(os.path.join(root, fname))
             if output:
-                key = lambda x: os.path.getmtime(_safe_path(x))
+                key = lambda x: getmtime(x)
```
gilles-duboscq: This is causing warnings in pylint. You can remove that local variable and use `key=getmtime` below.

jphalimi-intel:
I fixed the lint issue. thanks for spotting it.

gilles-duboscq:
Thanks, looks good now.

Issue #168, opened 2018-11-09
Python 3 support.


JornVernee:
Is python 3 support planned at all?

I was having a problem with paths exceeding max path length on Windows, and this is apparently fixed in python 3.6 (I worked around the problem by moving the suite directory closer to drive root)

I'm not a python expert, but I've seen applications supporting both python 2 and 3 before by using `import __future__`. Would that be a possible migration path?

dougxc:
We don't have any current plans for supporting python 3. However, we're open to anyone who wants to put together a plan and PR for adding python 3 support while ensuring mx will continue to work on 2.7.

JornVernee:
@dougxc I've looked into this once in the past. Of course Windows support wasn't as good as now, so I wasn't able to fully test my changes, but there were only 6 issues I ran into; 
1. print is a function in python 3, but you can get forward compatibility by using `import __future__`, and installing the `future` package in python 2.
2. python 3 doesn't have a `cmp` function, but there are easy workaround for this.
3. [some issue around OrderedDict](https://stackoverflow.com/a/39284955)
4. octal literals having a different syntax
5. results from `check_output` needing to be decoded into strings manually.
6. modules, types and functions being renamed.

I can see not much problems for 1-5, but 6 might be tricky to get to work with both versions.

I think 1 by far touches the most code, but it's also the most straight forward to do, as long as you don't mind using a print function instead of statement. I think that's a good place to start.

JornVernee:
@dougxc I have a patch to move to using `print` function in the `mx` repo, and it turned out there's no need to install an extra package after all, so clients don't need to update anything. Are you OK with submitting changes one by one instead of one big patch that adds python 3 support?

I've looked a little more into getting `mx` to work with python 2 & 3, but the changes I made in the past were changing some of the default encodings used, which meant that the XML parser couldn't recognize the encoding any more when running on python 2, since its went from ascii to unicode. (I haven't found a solution yet for that).

I think that with portable changes (like print function) we can get as close as possible to python 3 while still supporting python 2, and then if needed create a separate python 3 fork/branch to do the last few changes that are not compatible with python 2.

dougxc:
With adding python 3 support to mx, I think fined grained pull requests to the master branch is the best way to go to avoid regressions.

JornVernee:
Been looking into this some more today.

The train seems to stop at some point because the current version of astroid does not support new python 3 versions.

What is the reason for not updating to the latest version currently?

[...]

PR #172, opened 2018-11-12
Change mx to using print function, on the road to Python 3 support

JornVernee:
In a move towards supporting python 3, I have changed all print statements in the `mx` repo to the `print` function using the `__future__` import.

The print statement is not syntactically compatible with python 3, but the print function is.

See also: https://github.com/graalvm/mx/issues/168

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -29,6 +29,8 @@ r"""
 mx is a command line tool for managing the development of Java code organized as suites of projects.
 
 """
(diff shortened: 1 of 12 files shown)


JornVernee:
Also, maybe it's a good idea to add a code style check that blocks use of the print statement, which the advice to use the print function instead.

I'm not sure how this could be done though.

PR #179, opened 2018-11-29
[WIP] [Python 3] Remove use of types module and basestring

JornVernee:
Python 3.1 removes the builtin types from the `types.` module, so it's use in that regard is non-portable. This PR changes the use of `types` with builtin type names where appropriate.

Additionally, since `basestring` is no longer in Python 3, it's usage is exchanged for `str`. The difference is that `basestring` is a common super type for `unicode` and `str`, so this is technically more restrictive, but also portable. I don't think using `str` instead of `basestring` will matter in practice, since it's only being used in type checks with `isinstance` to check if something is a string, and almost all the code uses the `str` type, since it's the default for strings.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -44,7 +44,7 @@ try:
 except ImportError:
     from xml.etree.ElementTree import parse as etreeParse
 
-import os, errno, time, subprocess, shlex, types, StringIO, zipfile, signal, tempfile, platform
+import os, errno, time, subprocess, shlex, StringIO, zipfile, signal, tempfile, platform
 import __builtin__
 import textwrap
 import socket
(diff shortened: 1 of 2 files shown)


JornVernee:
Just did a force push since I realized a small mistake I made.

The `2to3` tool maps `types.StringType` to `bytes`, so that's what I used initially, but given the context it should probably just have been `str`, so I'm now used that instead. They are the same object in python 2 any ways, so it should only matter for python 3.

JornVernee:
Just as a heads up; This will probably conflict with https://github.com/graalvm/mx/pull/176 and https://github.com/graalvm/mx/pull/178 since they modify some of the same lines (I missed this earlier, sorry)

JornVernee:
Needs some other changes as well, will update once the dependencies are done

PR #190, opened 2019-04-08
Support overriding gcc/g++/ar from environment

Keruspe:
Most build system support this, and on some environment, raw "gcc" "g++" and "ar" might not be available

### DIFF
--- a/mx_native.py
+++ b/mx_native.py
@@ -342,7 +342,9 @@ class NinjaManifestGenerator(object):
             depfile = None
             deps = 'msvc'
         else:
-            command = '%s -MMD -MF $out.d $includes $cflags -c $in -o $out' % ('g++' if cxx else 'gcc')
+            c_compiler = os.getenv('CC', 'gcc')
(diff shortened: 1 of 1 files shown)


Keruspe:
OCA sent

Keruspe:
Is there really no way to do this online? It's really boring having to wait to get my hands on a printer/scanner just to resubmit because it was rejected the first time.

dougxc:
The OCA signing process is not under the control of the GraalVM teams.
I suspect (hope!) it will improve when OpenJDK moves to GitHub.

Keruspe:
Done

rschatz:
Do we still need this? I think the toolchains are now defined in `ninja-toolchain/*.ninja`, instead of hardcoded.

Looks like this code changed exactly in the time between signing the OCA and it getting processed by our bot. In particular this https://github.com/graalvm/mx/commit/403e97800ef898ccdf04b75a8783c5a123048fc4 looks related. @gilles-duboscq Is this doing the same thing? Or is this an orthogonal?

PR #195, opened 2019-06-10
Correct error log message post compiler daemon crash

neomatrix369:
Correcting the error message printed when the request sent to the compiler daemon fails: compiler daemon is called javac and not java. Also adding a space to separate the command from the arguments.

On the back of [the discussions](https://github.com/oracle/graal/issues/1243#issuecomment-500603776) on the reported issue https://github.com/oracle/graal/issues/1243

See output at https://circleci.com/gh/neomatrix369/awesome-graal/423 (search for `CalledProcessError`, we now see the correct message:

```
CalledProcessError: Command '/root/project/graal-jvmci-8/openjdk1.8.0-internal/linux-amd64/product/bin/javac -g -d /root/project/graal/compiler/mxbuild/src/org.graalvm.compiler.replacements.test/bin
[snipped]
```

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -4282,7 +4282,7 @@ class CompilerDaemon(Daemon):
         if retcode:
             if _opts.verbose:
                 if _opts.very_verbose:
-                    raise subprocess.CalledProcessError(retcode, self.jdk.java + ' '.join(compilerArgs))
+                    raise subprocess.CalledProcessError(retcode, self.jdk.javac + ' ' + ' '.join(compilerArgs))
                 else:
                     log('[exit code: ' + str(retcode) + ']')
             abort(retcode)
--- a/mx.py
+++ b/mx.py
@@ -4282,7 +4282,7 @@ class CompilerDaemon(Daemon):
         if retcode:
             if _opts.verbose:
                 if _opts.very_verbose:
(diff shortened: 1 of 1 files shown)


neomatrix369:
I have signed the OCA, see https://www.oracle.com/technetwork/community/oca-486395.html (search for neomatrix369, which part of the committer or author information is it using to search in the OCA database)?

neomatrix369:
I should have been registered under one of these email addresses, it's been a long time, would help if you can pls check the records for me.

mx.py
```diff
@@ -4282,7 +4282,7 @@ def compile(self, compilerArgs):
         if retcode:
             if _opts.verbose:
                 if _opts.very_verbose:
-                    raise subprocess.CalledProcessError(retcode, self.jdk.java + ' '.join(compilerArgs))
+                    raise subprocess.CalledProcessError(retcode, self.jdk.javac + ' ' + ' '.join(compilerArgs))
```
dougxc: Instead of throwing an exception, this should be:
```
    retcode = str(subprocess.CalledProcessError(retcode, self.jdk.javac + ' ' + ' '.join(compilerArgs)))
```

neomatrix369: Applied via 72083a4

Results can be seen at https://circleci.com/gh/neomatrix369/awesome-graal/426

PR #203, opened 2019-11-15
Upgrade to checkstyle 8.26

fniephaus:
and move LineLengthCheck outside of TreeWalker.

This is required to make all .checkstyle_checks.xml files compatible with checkstyle 8.24 and later.

See https://checkstyle.sourceforge.io/releasenotes.html#Release_8.24 and checkstyle/checkstyle#2116


fniephaus:
The mirror must also be updated with:
`https://lafo.ssw.uni-linz.ac.at/pub/graal-external-deps/checkstyle-8.26-all.jar`.

dougxc:
mx supports Checkstyle 8.36.1 since 294db340d301d5619c8c6c38ccef934a1947d249

PR #209, opened 2020-03-05
Add --jmods-dir

christhalinger:
This option allows us to build Graal with an exploded JDK like this:

```
mx --java-home=$(JDK_OUTPUTDIR) --jmods-dir=$(IMAGES_OUTPUTDIR)/jmods
```

This is necessary if someone wants to build a JDK image with an upstream Graal module.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -526,6 +526,7 @@ environment variables:
         self.add_argument('--version', action='store_true', help='print version and exit')
         self.add_argument('--mx-tests', action='store_true', help='load mxtests suite (mx debugging)')
         self.add_argument('--jdk', action='store', help='JDK to use for the "java" command', metavar='<tag:compliance>')
+        self.add_argument('--jmods-dir', action='store', help='path to built jmods (default JAVA_HOME/jmods)', metavar='<path>')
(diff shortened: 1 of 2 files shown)


dougxc:
Looks good to me and I'll integrate it. Can say more about how this helps build a JDK image with an upstream Graal module.

dougxc:
@christhalinger can you say more about how this helps build a JDK image with an upstream Graal module? Just curious if it sheds light on a more efficient way to do cross JDK and Graal development. I'd like to use an exploded JDK for development since it's much faster to build.

christhalinger:
It needs a bit of Makefile machinery but the idea is to build an exploded JDK, have the `mx` command to build Graal depend on `jmods` and then when running `make images` import the upstream modules.  I could paste the patch if that helps.

It would be possible to copy all generated class files into:

```
build/macosx-x86_64-normal-server-release/jdk/modules/jdk.internal.vm.compiler/
```

but I have not tried that.

christhalinger:
What would help would be a `--destination` option for `build`.  If we could put all the class files directly into the exploded module directory then everything would be much easier.  Even for me.

dougxc:
> I could paste the patch if that helps

That would be great.

dougxc:
> What would help would be a `--destination` option for `build`

I think that would be quite a significant change. Almost a redesign given how much mx depends on projects having their classes in separate directories.

PR #212, opened 2020-04-21
Make --dependencies/--projects and --only mutually exclusive

### LINKED ISSUE #211: --dependencies/--projects and --only appear to be mutually exclusive
If both `--dependencies` (or `--projects`) and `--only` are given to `mx build`, `--dependencies/--projects` will be ignored (https://github.com/graalvm/mx/blob/master/mx.py#L13512-L13520).
As a result these options should be made mutually exclusive.

zakkak:
Closes #211


dougxc:
Looks good - thanks! For future reference, there's no need to create an issue if you're submitting a pull request to address the issue. Just describe the issue in the pull request.

PR #214, opened 2020-04-29
Add `fetch-jdk` command

lazar-mitrovic:
Adds `mx fetch-labsjdk` command to `mx` which downloads JDK required for project building and outputs corresponding JAVA_HOME export command to stdout. 
```
$ mx fetch-jdk --help
usage: mx fetch-labsjdk [-h] [--java-distribution JAVA_DISTRIBUTION]
                        [--configuration CONFIGURATION] [--to TO]
                        [--alias ALIAS] [--keep-archive] [-q]
                        ...

optional arguments:
  -h, --help            show this help message and exit
  --java-distribution JAVA_DISTRIBUTION
                        JDK distribution which should be downloaded
  --configuration CONFIGURATION
                        location of configuration json file
  --to TO               location where JDK would be downloaded
  --alias ALIAS         name of symlink to JDK
  --keep-archive        keep downloaded JDK archive
  -q, --quiet           suppress logging output
```
If command is ran without `quiet` and `java-distribution` flags, interactive menu would be displayed.

Additional suites can register their own JDK distributions in their mx extensions, like:
```python
from jdk_distribution_parser import JdkDistribution

class LabsJDKEE(JdkDistribution):
    _name = "labsjdk-ee-11"
    def __init__(self, version):
        JdkDistribution.__init__(self, version)
        machine = mx.get_os() + '-' + mx.get_arch()
        self._filename = "labsjdk-{}-{}".format(version, machine)
        self._short_version = version
        self._archive = "{}.tar.gz".format(self._filename)
        self._url = "http://example.com/{}".format(self._archive)

JdkDistribution._DEFAULT_JDK = "labsjdk-ee-11"
```

### DIFF
--- a/jdk_distribution_parser.py
+++ b/jdk_distribution_parser.py
@@ -1,7 +1,7 @@
 #
 # ----------------------------------------------------------------------------------------------------
 #
-# Copyright (c) 2007, 2020, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
 # DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
 #
 # This code is free software; you can redistribute it and/or modify it
@@ -115,22 +115,6 @@ class OpenJDK8(JdkDistribution):
                     "{GITHUB_RELEASES}/{short_version}/{archive}"
                     ).format(GITHUB_URL=GITHUB_URL, GITHUB_RELEASES=GITHUB_RELEASES,
                     short_version=self._short_version, archive=self._archive)
-
-class OpenJDK11(JdkDistribution):
-    _name = "openjdk11"
-    def __init__(self, version):
-        JdkDistribution.__init__(self, version)
-        machine = self.get_machine()
-        self._short_version = version.replace("+", "_")
-        self._filename = "OpenJDK11U-jdk_{machine}_hotspot_{short_version}".format(short_version=self._short_version, machine=machine)
-        self._archive = "{}.tar.gz".format(self._filename)
-        self._url = ("{GITHUB_URL}/AdoptOpenJDK/openjdk11-binaries/"
-                    "{GITHUB_RELEASES}/jdk-{version}/{archive}"
-                    ).format(GITHUB_URL=GITHUB_URL, GITHUB_RELEASES=GITHUB_RELEASES,
-                    version=version, archive=self._archive)
-
-    def get_machine(self):
-        return mx.get_arch().replace("amd", "x") + "_" + mx.get_os().replace("darwin", "mac")
 
 class LabsJDK11CE(JdkDistribution):
     _name = "labsjdk-ce-11"
--- a/jdk_distribution_parser.py
+++ b/jdk_distribution_parser.py
@@ -24,7 +24,7 @@
 #
 # ----------------------------------------------------------------------------------------------------
 #
-import re, json
+import re
 import mx
 from os.path import join
 
@@ -41,13 +41,6 @@ class JdkDistribution(object):
             if dist._name == name:
                 dist(version)
                 return
-    @staticmethod
-    def parse_common_json(common_path):
-        with open(common_path) as common_file:
-            common_cfg = json.load(common_file)
-
-        for distribution in common_cfg["jdks"]:
-            JdkDistribution.parse(distribution, common_cfg["jdks"][distribution]["version"])
 
     @staticmethod
     def choose_dist(quiet=False):
--- a/jdk_distribution_parser.py
+++ b/jdk_distribution_parser.py
@@ -58,10 +58,9 @@ class JdkDistribution(object):
             name=dist.get_name().ljust(25), version=dist.get_version(), default=default))
             index += 1
         while True:
-            print("Select JDK>"), # pylint: disable=expression-not-assigned
             try:
                 try:
-                    choice = input()
+                    choice = input("Select JDK>")
                 except SyntaxError: # Empty line
                     choice = ""
 
@@ -91,7 +90,7 @@ class JdkDistribution(object):
     def get_filename(self):
         return self._filename
 
-    def get_archive(self):
+    def get_archive_name(self):
         return self._archive
 
     def get_url(self):
@@ -106,11 +105,11 @@ class JdkDistribution(object):
     def get_short_version(self):
         return self._short_version
 
-    def get_jdk_folder(self):
+    def get_folder_name(self):
         return "{}-{}".format(self.get_name(), self.get_short_version())
 
-    def get_full_jdk_path(self, jdk_path):
-        return join(jdk_path, self.get_jdk_folder())
+    def get_final_path(self, jdk_path):
+        return join(jdk_path, self.get_folder_name())
 
 class OpenJDK8(JdkDistribution):
     _name = "openjdk8"
--- a/jdk_distribution_parser.py
+++ b/jdk_distribution_parser.py
@@ -132,7 +132,7 @@ class OpenJDK11(JdkDistribution):
     def get_machine(self):
         return mx.get_arch().replace("amd", "x") + "_" + mx.get_os().replace("darwin", "mac")
 
-class LabsJDKCE(JdkDistribution):
+class LabsJDK11CE(JdkDistribution):
     _name = "labsjdk-ce-11"
     def __init__(self, version):
         JdkDistribution.__init__(self, version)
--- a/jdk_distribution_parser.py
+++ b/jdk_distribution_parser.py
@@ -40,7 +40,6 @@ class JdkDistribution(object):
         for dist in JdkDistribution.__subclasses__():
             if dist._name == name:
                 dist(version)
-                return
 
     @staticmethod
     def choose_dist(quiet=False):
@@ -51,7 +50,7 @@ class JdkDistribution(object):
         for dist in JdkDistribution._jdk_distributions:
             default = " " if dist.get_name() != JdkDistribution._DEFAULT_JDK else "*"
             print("[{index}]{default} {name} | {version}".format(index=index,
-            name=dist.get_name().ljust(15), version=dist.get_version(), default=default))
+            name=dist.get_name().ljust(25), version=dist.get_version(), default=default))
             index += 1
         while True:
             print("Select JDK>"), # pylint: disable=expression-not-assigned
@@ -104,10 +103,26 @@ class OpenJDK8(JdkDistribution):
         self._filename = "openjdk-{}-{}".format(version, machine)
         self._short_version = re.sub(r".*(jvmci.*)", "\\1", version)
         self._archive = "{}.tar.gz".format(self._filename)
-        self._url = ("{GITHUB_URL}/graalvm/openjdk8-jvmci-builder/"
+        self._url = ("{GITHUB_URL}/graalvm/graal-jvmci-8/"
                     "{GITHUB_RELEASES}/{short_version}/{archive}"
                     ).format(GITHUB_URL=GITHUB_URL, GITHUB_RELEASES=GITHUB_RELEASES,
                     short_version=self._short_version, archive=self._archive)
+
+class OpenJDK8Debug(JdkDistribution):
+    _name = "openjdk8"
+    def __init__(self, version):
+        JdkDistribution.__init__(self, version)
+        machine = mx.get_os() + '-' + mx.get_arch()
+        self._filename = "openjdk-{}-fastdebug-{}".format(version, machine)
+        self._short_version = re.sub(r".*(jvmci.*)", "\\1", version)
+        self._archive = "{}.tar.gz".format(self._filename)
+        self._url = ("{GITHUB_URL}/graalvm/graal-jvmci-8/"
+                    "{GITHUB_RELEASES}/{short_version}/{archive}"
+                    ).format(GITHUB_URL=GITHUB_URL, GITHUB_RELEASES=GITHUB_RELEASES,
+                    short_version=self._short_version, archive=self._archive)
+
+    def get_name(self):
+        return self._name + "-fastdebug"
 
 class LabsJDK11CE(JdkDistribution):
     _name = "labsjdk-ce-11"
@@ -121,3 +136,19 @@ class LabsJDK11CE(JdkDistribution):
                     "{GITHUB_RELEASES}/{short_version}/{archive}"
                     ).format(GITHUB_URL=GITHUB_URL, GITHUB_RELEASES=GITHUB_RELEASES,
                     short_version=self._short_version, archive=self._archive)
+
+class LabsJDK11CEDebug(JdkDistribution):
+    _name = "labsjdk-ce-11"
+    def __init__(self, version):
+        JdkDistribution.__init__(self, version)
+        machine = mx.get_os() + '-' + mx.get_arch()
+        self._filename = "labsjdk-{}-debug-{}".format(version, machine)
+        self._short_version = re.sub(r".*(jvmci.*)", "\\1", version)
+        self._archive = "{}.tar.gz".format(self._filename)
+        self._url = ("{GITHUB_URL}/graalvm/labs-openjdk-11/"
+                    "{GITHUB_RELEASES}/{short_version}/{archive}"
+                    ).format(GITHUB_URL=GITHUB_URL, GITHUB_RELEASES=GITHUB_RELEASES,
+                    short_version=self._short_version, archive=self._archive)
+
+    def get_name(self):
+        return self._name + "-debug"
--- a/jdk_distribution_parser.py
+++ b/jdk_distribution_parser.py
@@ -47,18 +47,35 @@ class JdkDistribution(object):
             return JdkDistribution.by_name(JdkDistribution._DEFAULT_JDK)
 
         index = 1
+        default_choice = 1
         for dist in JdkDistribution._jdk_distributions:
-            default = " " if dist.get_name() != JdkDistribution._DEFAULT_JDK else "*"
+            if dist.get_name() == JdkDistribution._DEFAULT_JDK:
+                default_choice = index
+                default = "*"
+            else:
+                default = " "
             print("[{index}]{default} {name} | {version}".format(index=index,
             name=dist.get_name().ljust(25), version=dist.get_version(), default=default))
             index += 1
         while True:
             print("Select JDK>"), # pylint: disable=expression-not-assigned
             try:
-                index = int(input()) - 1
+                try:
+                    choice = input()
+                except SyntaxError: # Empty line
+                    choice = ""
+
+                if choice == "":
+                    index = default_choice - 1
+                else:
+                    index = int(choice) - 1
+
+                if index < 0:
+                    raise IndexError
+
                 return JdkDistribution._jdk_distributions[index]
             except (SyntaxError, NameError, IndexError):
-                pass
+                mx.warn("Invalid selection!")
 
     @staticmethod
     def by_name(name):
--- a/jdk_distribution_parser.py
+++ b/jdk_distribution_parser.py
@@ -44,7 +44,7 @@ class JdkDistribution(object):
     @staticmethod
     def parse_common_json(common_path):
         with open(common_path) as common_file:
-                common_cfg = json.load(common_file)
+            common_cfg = json.load(common_file)
 
         for distribution in common_cfg["jdks"]:
             JdkDistribution.parse(distribution, common_cfg["jdks"][distribution]["version"])
@@ -52,16 +52,16 @@ class JdkDistribution(object):
     @staticmethod
     def choose_dist(quiet=False):
         if quiet:
-           return JdkDistribution.by_name(JdkDistribution._DEFAULT_JDK)
- 
+            return JdkDistribution.by_name(JdkDistribution._DEFAULT_JDK)
+
         index = 1
         for dist in JdkDistribution._jdk_distributions:
             default = " " if dist.get_name() != JdkDistribution._DEFAULT_JDK else "*"
-            print("[{index}]{default} {name} | {version}".format(index=index, 
+            print("[{index}]{default} {name} | {version}".format(index=index,
             name=dist.get_name().ljust(15), version=dist.get_version(), default=default))
             index += 1
         while True:
-            print("Select JDK>"),
+            print("Select JDK>"), # pylint: disable=expression-not-assigned
             try:
                 index = int(input()) - 1
                 return JdkDistribution._jdk_distributions[index]
@@ -91,9 +91,6 @@ class JdkDistribution(object):
     def get_name(self):
         return self._name
 
-    def get_filename(self):
-        return self._filename
-    
     def get_version(self):
         return self._version
 
@@ -116,9 +113,8 @@ class OpenJDK8(JdkDistribution):
         self._archive = "{}.tar.gz".format(self._filename)
         self._url = ("{GITHUB_URL}/graalvm/openjdk8-jvmci-builder/"
                     "{GITHUB_RELEASES}/{short_version}/{archive}"
-                    ).format(GITHUB_URL=GITHUB_URL, GITHUB_RELEASES=GITHUB_RELEASES, 
-                    version=version, short_version=self._short_version, archive=self._archive, 
-                    machine=machine)
+                    ).format(GITHUB_URL=GITHUB_URL, GITHUB_RELEASES=GITHUB_RELEASES,
+                    short_version=self._short_version, archive=self._archive)
 
 class OpenJDK11(JdkDistribution):
     _name = "openjdk11"
@@ -130,9 +126,8 @@ class OpenJDK11(JdkDistribution):
         self._archive = "{}.tar.gz".format(self._filename)
         self._url = ("{GITHUB_URL}/AdoptOpenJDK/openjdk11-binaries/"
                     "{GITHUB_RELEASES}/jdk-{version}/{archive}"
-                    ).format(GITHUB_URL=GITHUB_URL, GITHUB_RELEASES=GITHUB_RELEASES, 
-                    version=version, short_version=self._short_version, archive=self._archive, 
-                    machine=machine)
+                    ).format(GITHUB_URL=GITHUB_URL, GITHUB_RELEASES=GITHUB_RELEASES,
+                    version=version, archive=self._archive)
 
     def get_machine(self):
         return mx.get_arch().replace("amd", "x") + "_" + mx.get_os().replace("darwin", "mac")
@@ -147,6 +142,5 @@ class LabsJDKCE(JdkDistribution):
         self._archive = "{}.tar.gz".format(self._filename)
         self._url = ("{GITHUB_URL}/graalvm/labs-openjdk-11/"
                     "{GITHUB_RELEASES}/{short_version}/{archive}"
-                    ).format(GITHUB_URL=GITHUB_URL, GITHUB_RELEASES=GITHUB_RELEASES, 
-                    version=version, short_version=self._short_version, archive=self._archive, 
-                    machine=machine)
+                    ).format(GITHUB_URL=GITHUB_URL, GITHUB_RELEASES=GITHUB_RELEASES,
+                    short_version=self._short_version, archive=self._archive)
--- a/jdk_distribution_parser.py
+++ b/jdk_distribution_parser.py
@@ -1,0 +1,152 @@
+#
+# ----------------------------------------------------------------------------------------------------
+#
+# Copyright (c) 2007, 2020, Oracle and/or its affiliates. All rights reserved.
+# DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+#
+# This code is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 only, as
+# published by the Free Software Foundation.
+#
+# This code is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+# version 2 for more details (a copy is included in the LICENSE file that
+# accompanied this code).
+#
+# You should have received a copy of the GNU General Public License version
+# 2 along with this work; if not, write to the Free Software Foundation,
+# Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+# or visit www.oracle.com if you need additional information or have any
+# questions.
+#
+# ----------------------------------------------------------------------------------------------------
+#
+import re, json
+import mx
+from os.path import join
+
+GITHUB_URL = "https://github.com"
+GITHUB_RELEASES = "releases/download"
+
+class JdkDistribution(object):
+    _jdk_distributions = []
+    _DEFAULT_JDK = "labsjdk-ce-11"
+
+    @staticmethod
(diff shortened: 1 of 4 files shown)


mx_fetchjdk.py
```diff
@@ -0,0 +1,147 @@
...
+
+    if not args["quiet"]:
+        print("Run the following to set JAVA_HOME in your shell:")
+
+    print('export JAVA_HOME={}'.format(abspath(full_jdk_path)))
+
+
+def _parse_fetchjdk_settings(args):
+    parser = ArgumentParser(prog='mx fetch-labsjdk')
+    parser.add_argument('--java-distribution', action='store', help='JDK distribution which should be downloaded')
```
dougxc, line 95: It helps to give examples of the kind of values an argument accepts:
```
    parser.add_argument('--java-distribution', action='store', help='JDK distribution which should be downloaded (e.g. "labsjdk-ce-11" or "openjdk8")')
```

mx_fetchjdk.py
```diff
@@ -0,0 +1,147 @@
...
+    parser.add_argument('--configuration', action='store', help='location of configuration json file')
+    parser.add_argument('--to', action='store', help='location where JDK would be downloaded')
+    parser.add_argument('--alias', action='store', help='name of symlink to JDK')
+    parser.add_argument('--keep-archive', action='store_true', help='keep downloaded JDK archive')
+    parser.add_argument('-q', '--quiet', action='store_true', help='suppress logging output')
+    parser.add_argument('remainder', nargs=REMAINDER, metavar='...')
+    args = parser.parse_args(args)
+
+    settings = {}
+    settings["jdk_path"] = join('bin', 'jdks')
```
dougxc, line 105: This should be `os.path.abspath(join('bin', 'jdks'))`.

mx_fetchjdk.py
```diff
@@ -0,0 +1,147 @@
...
+    if not args["quiet"]:
+        print("Run the following to set JAVA_HOME in your shell:")
+
+    print('export JAVA_HOME={}'.format(abspath(full_jdk_path)))
+
+
+def _parse_fetchjdk_settings(args):
+    parser = ArgumentParser(prog='mx fetch-labsjdk')
+    parser.add_argument('--java-distribution', action='store', help='JDK distribution which should be downloaded')
+    parser.add_argument('--configuration', action='store', help='location of configuration json file')
```
dougxc, line 96: Both the `--configuration` and `--to` arguments should mention their default values in their help text:
```
    parser.add_argument('--configuration', action='store', help='location of configuration json file (default: ' + common_location + ')')
    parser.add_argument('--to', action='store', help='location where JDK would be downloaded (default: ' + settings["jdk_path"] + ')')
```

Of course, this means moving the computation of the defaults above the initialization of `parser`.

mx_fetchjdk.py
```diff
@@ -0,0 +1,147 @@
...
+@optional_suite_context
+def fetch_jdk(args):
+    """fetches required JDK version
+
+    If called without --quiet flag, menu will be printed for available JDK selection.
+    """
+    args = _parse_fetchjdk_settings(args)
+
+    distribution = args["java_distribution"]
+    jdk_path = args["jdk_path"]
```
dougxc, line 48: You should check that the user can write to `jdk_path` and abort with a message about having to re-run with `sudo`. Unfortunately we cannot use 3rd party libs in mx such as [elevate](https://pypi.org/project/elevate/).

mx_fetchjdk.py
```diff
@@ -0,0 +1,147 @@
...
+    jdk_folder = distribution.get_jdk_folder()
+    full_jdk_path = distribution.get_full_jdk_path(jdk_path)
+    jdk_url = mx_urlrewrites.rewriteurl(distribution.get_url())
+    jdk_archive = distribution.get_archive()
+
+    if not exists(full_jdk_path):
```
dougxc, line 54: Don't you need an `else` statement if the JDK already exists locally?

mx.py
```diff
@@ -749,7 +749,7 @@ def suite_context_free(func):
     """
     Decorator for commands that don't need a primary suite.
     """
-    _suite_context_free.append(func.__name__)
+    _suite_context_free.append(func.__name__.replace("_", "-"))
```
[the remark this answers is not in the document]
lazar-mitrovic: This decorator adds _function name_ to list of _commands_. 
Since python doesn't support function names with `-` character, functions that implement  commands with `-` are named with underscore. 

Without this change all such commands wouldn't be registered as `suite_context_free` commands (`quiet_run` is the only one I'm able to find atm).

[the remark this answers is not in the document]
dougxc: Won't this be a problem if someone defines a command with an '_' in its name?

jdk_distribution_parser.py
```diff
@@ -0,0 +1,146 @@
+#
+# ----------------------------------------------------------------------------------------------------
+#
+# Copyright (c) 2007, 2020, Oracle and/or its affiliates. All rights reserved.
```
vjovanov, line 4: 2020

mx_fetchjdk.py
```diff
@@ -0,0 +1,161 @@
+#
+# ----------------------------------------------------------------------------------------------------
+#
+# Copyright (c) 2007, 2020, Oracle and/or its affiliates. All rights reserved.
```
vjovanov, line 4: 2020

jdk_distribution_parser.py
```diff
@@ -0,0 +1,146 @@
...
+        self._archive = "{}.tar.gz".format(self._filename)
+        self._url = ("{GITHUB_URL}/AdoptOpenJDK/openjdk11-binaries/"
+                    "{GITHUB_RELEASES}/jdk-{version}/{archive}"
+                    ).format(GITHUB_URL=GITHUB_URL, GITHUB_RELEASES=GITHUB_RELEASES,
+                    version=version, archive=self._archive)
+
+    def get_machine(self):
+        return mx.get_arch().replace("amd", "x") + "_" + mx.get_os().replace("darwin", "mac")
+
+class LabsJDK11CE(JdkDistribution):
```
vjovanov, line 135: This is quite graal-specific. Can we put this into the `sdk` project?

mx_fetchjdk.py
```diff
@@ -0,0 +1,161 @@
...
+    if len(jdk_paths) > 0:
+        settings["jdk_path"] = jdk_paths[0]
+        found_jdk_path = True
+    else:
+        found_jdk_path = False
+
+    common_location = join(_mx_home, 'common.json')
+
+    parser = ArgumentParser(prog='mx fetch-labsjdk')
+    parser.add_argument('--java-distribution', action='store', help='JDK distribution which should be downloaded (e.g. "labsjdk-ce-11" or "openjdk8")')
```
vjovanov, line 109: which -> that

vjovanov: e.g.,

mx.py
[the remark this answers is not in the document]
lazar-mitrovic: My idea was to rely on developers using consistent naming style. 
Only other solution I see would be to have decorator parameter which would toggle this replacement on case-by-case basis.

mx_fetchjdk.py
lazar-mitrovic: If JDK exists then only `alias` is created (if corresponding argument was set), and `export JAVA_HOME` is printed.

mx.py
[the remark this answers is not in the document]
dougxc: It's fine to forbid command names containing '_' but then you need to enforce this otherwise someone is going to be very confused when they define a suite context free command named `my_command` only to have it fail in the absence of a suite.

mx_commands.py
```diff
@@ -114,6 +114,8 @@ def add_commands(self, new_commands):
 class MxCommand(object):
 
     def __init__(self, mx_commands, command_function, suite_name, command, usage_msg='', doc_function=None, props=None):
+        if '_' in command:
+            mx.abort("MX Command '{}' contains '_' in its name instead of '-'".format(command))
```
dougxc, line 118: A slight better error message:
```
"An mx command cannot contain '_' in its name: {}. Use '-' instead.".format(command)
```

lazar-mitrovic: Should be fixed now

mx_fetchjdk.py
dougxc: Then we should at least print a comment about the JDK already being downloaded in non-quiet mode. It may surprise people how fast the download occurs otherwise ;-)

vjovanov:
```$ mx fetch-jdk
  File "/Users/vjovanov/dev/mx/mx.py", line 19523, in <module>
    main()
  File "/Users/vjovanov/dev/mx/mx.py", line 19504, in main
    retcode = c(command_args)
  File "/Users/vjovanov/dev/mx/mx_commands.py", line 147, in __call__
    return self.command_function(*args, **kwargs)
  File "/Users/vjovanov/dev/mx/mx_fetchjdk.py", line 45, in fetch_jdk
    args = _parse_fetchjdk_settings(args)
  File "/Users/vjovanov/dev/mx/mx_fetchjdk.py", line 103, in _parse_fetchjdk_settings
    settings["jdk-path"] = jdk_paths[0]
TypeError: 'set' object does not support indexing```

File: mx_fetchjdk.py [hunk 12]
```diff
@@ -0,0 +1,165 @@
...
+    jdk_paths = find_system_jdks()
+    if len(jdk_paths) > 0:
+        settings["jdk-path"] = jdk_paths[0]
+        found_jdk_path = True
+    else:
+        found_jdk_path = False
+
+    common_location = join(_mx_home, 'common.json')
+
+    parser = ArgumentParser(prog='mx fetch-labsjdk')
```
vjovanov, line 110: `mx fetch-jdk`

vjovanov:
After fixing the issue, the script tries to install the JDK in a wrong place:
```
(master ☡=)$ mx fetch-jdk
Path '/Library/Java/JavaVirtualMachines/jdk1.8.0_221.jdk/Contents/Home' is not writable. 
Rerun command with elevated privileges, or choose different JDK download location.
```

mx_fetchjdk.py
```diff
@@ -0,0 +1,165 @@
...
+            curr_full_jdk_path = untar._getnames(tar_file)[0]
+
+        untar.extract(jdk_path)
+        if not args["keep-archive"]:
+            os.remove(join(jdk_path, jdk_archive))
+        os.rename(join(jdk_path, curr_full_jdk_path), full_jdk_path)
+    elif not args["quiet"]:
+        mx.warn("Requested JDK was already present")
+
+    if mx.get_os() == 'darwin':
```
vjovanov, line 72: I would use `is_darwin`.

mx_fetchjdk.py
```diff
@@ -0,0 +1,165 @@
...
+        full_jdk_path = join(full_jdk_path, 'Contents', 'Home')
+
+    if "alias" in args:
+        alias_full_path = join(jdk_path, args["alias"])
```
vjovanov, line 76: `Contents/Home` is missing here.

mx_fetchjdk.py
vjovanov: I need an option on darwin to strip `Contents/Home`.

mx_fetchjdk.py
```diff
@@ -0,0 +1,165 @@
...
+        if not (mx.is_windows() or mx.is_cygwin()):
+            os.symlink(jdk_folder, alias_full_path)
+        else:
+            copytree(full_jdk_path, alias_full_path, symlinks=True) # fallback for windows
+        full_jdk_path = alias_full_path
+
+    if not args["quiet"]:
+        print("Run the following to set JAVA_HOME in your shell:")
+
+    print('export JAVA_HOME={}'.format(abspath(full_jdk_path)))
```
vjovanov, line 92: `Contents/Home` is missing here.

mx_fetchjdk.py
```diff
@@ -0,0 +1,165 @@
...
+            common_location = join(os.getcwd(), 'common.json') # Fallback to same folder
+            if not exists(common_location):
+                common_location = join(_mx_home, 'common.json') # Fallback to mx
+            mx.warn("Selected `{}` as configuration location, since no location is provided".format(common_location))
+    if not exists(common_location):
+        mx.abort("Configuration file doesn't exist")
+
+    JdkDistribution.parse_common_json(common_location)
+
+    if args.java_distribution is not None:
```
vjovanov, line 154: I think it there would be less redundancy if all of this was done by the options parser.

mx_fetchjdk.py
```diff
@@ -0,0 +1,165 @@
...
+        settings["java-distribution"] = JdkDistribution.by_name(args.java_distribution)
+    else:
+        settings["java-distribution"] = JdkDistribution.choose_dist(settings["quiet"])
+
+    if args.keep_archive is not None:
+        settings["keep-archive"] = args.keep_archive
```
vjovanov, line 160: This setting requires 4 lines instead of just one.

mx_fetchjdk.py
```diff
@@ -0,0 +1,165 @@
...
+        untar = mx.TarExtractor(join(jdk_path, jdk_archive))
+
+        if not args["quiet"]:
+            print("Installing...")
+
+        with untar._open() as tar_file:
+            curr_full_jdk_path = untar._getnames(tar_file)[0]
+
+        untar.extract(jdk_path)
+        if not args["keep-archive"]:
```
vjovanov, line 66: This should be in the try-catch. If something fails the archive is kept and it actually breaks the consecutive download.

vjovanov:
With no JDK I am getting this: 
  Path in JAVA_HOME is not pointing to a JDK...

That makes no sense for a Python build tool. Can we sink this check to the place where it is actually needed?

vjovanov:
@lazar-mitrovic  Can we go through the `mx` cache for fetching the JDK? @dougxc can we upload the digest together with the JDK so this script can check it? 
Also, it would be nice to upload the debug version of each JDK.

dougxc:
> @dougxc can we upload the digest together with the JDK so this script can check it?
> Also, it would be nice to upload the debug version of each JDK.

The debug versions and sha1 digests for the latest jvmci releases are at:
https://github.com/graalvm/graal-jvmci-8/releases/tag/jvmci-20.1-b02
https://github.com/graalvm/labs-openjdk-11/releases/tag/jvmci-20.1-b02

mx.py
```diff
@@ -750,6 +750,8 @@ def suite_context_free(func):
     Decorator for commands that don't need a primary suite.
     """
     _suite_context_free.append(func.__name__)
+    if '_' in func.__name__:
```
vjovanov, line 753: We should make this a function to avoid repetition.

dougxc:
I just pulled this branch and the first thing I did was:
```
dnsimon@klar-4 ~/mx (lazar-mitrovic-fetch-jdk)> mx fetch-jdk
Path '/Library/Java/JavaVirtualMachines/jdk-13.jdk/Contents/Home' is not writable.
Rerun command with elevated privileges, or choose different JDK download location.
```
That error is too eager since I may not select an action that writes to that location.

mx_fetchjdk.py
```diff
@@ -0,0 +1,197 @@
...
+
+def _parse_fetchjdk_settings(args):
+    settings = {}
+    settings["quiet"] = False
+    settings["keep-archive"] = False
+    settings["jdk-path"] = abspath(join('bin', 'jdks'))
+
+    jdk_paths = find_system_jdks()
+    if len(jdk_paths) > 0:
+        settings["jdk-path"] = next(iter(jdk_paths))
```
dougxc, line 122: This makes the default installation location:
a) non-deterministic (`find_system_jdks()` returns a set)
b) the path of `JAVA_HOME` for an installed JDK where as the `--to` argument should be the base directory into which the JDK is installed. That is, `/Library/Java/JavaVirtualMachines` on macOS, `/usr/lib/jvm` on Linux, `/usr/jdk/instances` on Solaris and `C:\Program Files\Java` on Windows.

mx_fetchjdk.py
```diff
@@ -0,0 +1,197 @@
...
+        found_jdk_path = True
+    else:
+        found_jdk_path = False
+
+    common_location = join(_mx_home, 'common.json')
+
+    parser = ArgumentParser(prog='mx fetch-jdk')
+    parser.add_argument('--java-distribution', action='store', help='JDK distribution that should be downloaded (e.g., "labsjdk-ce-11" or "openjdk8")')
+    parser.add_argument('--configuration', action='store', help='location of configuration json file (default: \'{}\')'.format(common_location))
+    parser.add_argument('--to', action='store', help='location where JDK would be downloaded (default: \'{}\')'.format(settings["jdk-path"]))
```
dougxc, line 132: As mentioned above, the `--to` argument is for specifying the directory in which the JDK is to be installed, not the `JAVA_HOME` directory of the installed JDK.

dougxc:
```
dnsimon@klar-4 ~/mx (lazar-mitrovic-fetch-jdk)> sudo mx fetch-jdk
[1]* labsjdk-ce-11             | ce-11.0.7+10-jvmci-20.1-b02
[2]  labsjdk-ce-11-debug       | ce-11.0.7+10-jvmci-20.1-b02
[3]  openjdk8                  | 8u252+09-jvmci-20.1-b02
[4]  openjdk8-fastdebug        | 8u252+09-jvmci-20.1-b02
Select JDK>
Select JDK>
Select JDK>
```

That is, it seems as though `labsjdk-ce-11` is the default choice (or does the "*" means something else?) but when I don't make a choice, I go straight back to the prompt.

dougxc:
```
dnsimon@klar-4 ~/mx (lazar-mitrovic-fetch-jdk)> sudo mx fetch-jdk
[1]* labsjdk-ce-11             | ce-11.0.7+10-jvmci-20.1-b02
[2]  labsjdk-ce-11-debug       | ce-11.0.7+10-jvmci-20.1-b02
[3]  openjdk8                  | 8u252+09-jvmci-20.1-b02
[4]  openjdk8-fastdebug        | 8u252+09-jvmci-20.1-b02
Select JDK>
Select JDK>
Select JDK> 2
Downloading labsjdk-ce-11-debug-jvmci-20.1-b02 from ['https://github.com/graalvm/labs-openjdk-11/releases/download/jvmci-20.1-b02/labsjdk-ce-11.0.7+10-jvmci-20.1-b02-debug-darwin-amd64.tar.gz']
 189374490 bytes (100%)
Installing...
Run the following to set JAVA_HOME in your shell:
export JAVA_HOME=/Library/Java/JavaVirtualMachines/jdk-11.0.5.0.50.jdk/Contents/Home/labsjdk-ce-11-debug-jvmci-20.1-b02
```

1. Instead of `Installing...`, it should be `Installing labsjdk-ce-11-debug to /Library/Java/JavaVirtualMachines/labsjdk-ce-11-debug-jvmci-20.1-b02 ...`. Always err on the side of being more informative. I would even such a confirmation step:
 ```
Install labsjdk-ce-11-debug to /Library/Java/JavaVirtualMachines/labsjdk-ce-11-debug-jvmci-20.1-b02 [Yn]?
y
Installing labsjdk-ce-11-debug to /Library/Java/JavaVirtualMachines/labsjdk-ce-11-debug-jvmci-20.1-b02 ...
 ```
2. The install location is wrong. Instead of `/Library/Java/JavaVirtualMachines/jdk-11.0.5.0.50.jdk/Contents/Home/labsjdk-ce-11-debug-jvmci-20.1-b02` it should be `/Library/Java/JavaVirtualMachines/labsjdk-ce-11-debug-jvmci-20.1-b02` and `JAVA_HOME` should be `/Library/Java/JavaVirtualMachines/labsjdk-ce-11-debug-jvmci-20.1-b02/Contents/Home`.
3. The .sha1 digest file is left lying around:
```
dnsimon@klar-4 ~/mx (lazar-mitrovic-fetch-jdk)> ls /Library/Java/JavaVirtualMachines/jdk-11.0.5.0.50.jdk/Contents/Home | grep labsjdk-ce-11
labsjdk-ce-11-debug-jvmci-20.1-b02
labsjdk-ce-11.0.7+10-jvmci-20.1-b02-debug-darwin-amd64.tar.gz.sha1
```

jdk_distribution_parser.py
```diff
@@ -0,0 +1,171 @@
...
+            if dist.get_name() == JdkDistribution._DEFAULT_JDK:
+                default_choice = index
+                default = "*"
+            else:
+                default = " "
+            print("[{index}]{default} {name} | {version}".format(index=index,
+            name=dist.get_name().ljust(25), version=dist.get_version(), default=default))
+            index += 1
+        while True:
+            print("Select JDK>"), # pylint: disable=expression-not-assigned
```
dougxc, line 61: `input` should be used here instead of `print` so that `Select JDK>` is a prompt on the same line that the user provides their input.

mx_fetchjdk.py
```diff
@@ -0,0 +1,213 @@
...
+    parser.add_argument('remainder', nargs=REMAINDER, metavar='...')
+    args = parser.parse_args(args)
+
+    if args.quiet is not None:
+        settings["quiet"] = args.quiet
+
+    if args.to is not None:
+        settings["jdk-path"] = args.to
+    elif not settings["quiet"]:
+        mx.warn("JDK location hasn't been specified. Using {}".format(settings["jdk-path"]))
```
dougxc, line 146: This warning is distracting. The user will see the JDK location once they've made their selection.

mx_fetchjdk.py
```diff
@@ -0,0 +1,213 @@
...
+
+    if not check_access_jdk_path(settings["jdk-path"]):
+        warning_msg = "Path '{}' is not writable (try running with elevated privileges). ".format(settings["jdk-path"])
+        settings["jdk-path"] = abspath(join("bin", "jdks"))
+        warning_msg += os.linesep + "Trying to fall back to {}".format(settings["jdk-path"])
```
dougxc, line 151: Don't fall back. If the user does not want to install in the system standard location, then they must use the `--to` option. The preceding warning about the system standard location not being writable should be converted to an abort:
```
mx.abort("JDK installation directory {} is not writeable. Either re-run with elevated privileges (e.g. sudo) or specify a writeable directory with the --to option.")
```

mx_fetchjdk.py
```diff
@@ -0,0 +1,213 @@
...
+    if not exists(full_jdk_path):
+        if not args["quiet"]:
+            print("Fetching {} archive...".format(jdk_artifact))
+
+        archive_location = join(jdk_path, jdk_archive)
+        mx._opts.no_download_progress = args["quiet"]
+        try:
+            sha1_hash = mx._hashFromUrl(jdk_url_sha)
+        except:
+            mx.abort("Error fetching sha1 hash, check your download location")
```
dougxc, line 66: In error handling, be clearer about what was tried and what might be the problem.
```
    except Exception as e:
            mx._suggest_http_proxy_error(e)
            mx.abort("Error fetching sha1 hash from {}: {}".format(jdk_url_sha , e))
```

mx_fetchjdk.py
```diff
@@ -0,0 +1,213 @@
...
+    jdk_url = mx_urlrewrites.rewriteurl(distribution.get_url())
+    jdk_url_sha = jdk_url + ".sha1"
+    jdk_archive = distribution.get_archive()
+
+    if not args["quiet"] and not mx.ask_yes_no("Install {} to {}".format(jdk_artifact, jdk_path), default='y'):
+        mx.abort("JDK installation canceled")
+
+    if not exists(full_jdk_path):
+        if not args["quiet"]:
+            print("Fetching {} archive...".format(jdk_artifact))
```
dougxc, line 59:
```
            print("Fetching {} archive from {} ...".format(jdk_artifact, jdk_url))
```

mx_fetchjdk.py
```diff
@@ -0,0 +1,213 @@
...
+
+        archive_location = join(jdk_path, jdk_archive)
+        mx._opts.no_download_progress = args["quiet"]
+        try:
+            sha1_hash = mx._hashFromUrl(jdk_url_sha)
```
dougxc, line 64: Seems like `sha1_hash` is of type `bytes` which causes a problem when running with Python3:
```
[1]  openjdk8                  | 8u252+09-jvmci-20.1-b02
[2]  openjdk8-fastdebug        | 8u252+09-jvmci-20.1-b02
[3]* labsjdk-ce-11             | ce-11.0.7+10-jvmci-20.1-b02
[4]  labsjdk-ce-11-debug       | ce-11.0.7+10-jvmci-20.1-b02
Select JDK>
Install labsjdk-ce-11-jvmci-20.1-b02 to /Users/dnsimon/mx/bin/jdks? [Yn]:
Fetching labsjdk-ce-11-jvmci-20.1-b02 archive...
  File "/Users/dnsimon/mx/mx.py", line 19529, in <module>
    main()
  File "/Users/dnsimon/mx/mx.py", line 19510, in main
    retcode = c(command_args)
  File "/Users/dnsimon/mx/mx_commands.py", line 147, in __call__
    return self.command_function(*args, **kwargs)
  File "/Users/dnsimon/mx/mx_fetchjdk.py", line 69, in fetch_jdk
    mx.download_file_with_sha1(jdk_artifact, archive_location, [jdk_url], sha1_hash, archive_location + '.sha1', resolve=True, mustExist=True, sources=False)
  File "/Users/dnsimon/mx/mx.py", line 3499, in download_file_with_sha1
    cachePath = _get_path_in_cache(name, sha1, urls, sources=sources)
  File "/Users/dnsimon/mx/mx.py", line 168, in _get_path_in_cache
    return join(_cache_dir(), name + '_' + sha1 + ('.dir' if not ext else ''), filename)
TypeError: can only concatenate str (not "bytes") to str
```

mx_fetchjdk.py
```diff
@@ -0,0 +1,213 @@
...
+        except:
+            mx.abort("Error fetching sha1 hash, check your download location")
+        if not exists(archive_location) or not args["keep-archive"]:
+            mx.download_file_with_sha1(jdk_artifact, archive_location, [jdk_url], sha1_hash, archive_location + '.sha1', resolve=True, mustExist=True, sources=False)
+        untar = mx.TarExtractor(archive_location)
+
+        if not args["quiet"]:
+            print("Installing {} to {}...".format(jdk_artifact, jdk_path))
```
dougxc, line 72: Move this to a point where `curr_full_jdk_path` has been defined and change it to:
```
            print("Installing {} to {} ...".format(jdk_artifact, curr_full_jdk_path))
```

dougxc:
```
...
Installing labsjdk-ce-11-jvmci-20.1-b02 to /Users/dnsimon/mx/bin/jdks...
Run the following to set JAVA_HOME in your shell:
export JAVA_HOME=/Users/dnsimon/mx/bin/jdks/labsjdk-ce-11-jvmci-20.1-b02
dnsimon@klar-4 ~/mx (lazar-mitrovic-fetch-jdk)> ls bin/jdks/
labsjdk-ce-11-jvmci-20.1-b02
dnsimon@klar-4 ~/mx (lazar-mitrovic-fetch-jdk)> ls bin/jdks/labsjdk-ce-11-jvmci-20.1-b02/
bin     conf    include jmods   legal   lib     release
dnsimon@klar-4 ~/mx (lazar-mitrovic-fetch-jdk)>
```
What happened to `Contents/Home`? This layout should be preserved on macOS.

jdk_distribution_parser.py
```diff
@@ -0,0 +1,146 @@
...
+        self._archive = "{}.tar.gz".format(self._filename)
+        self._url = ("{GITHUB_URL}/AdoptOpenJDK/openjdk11-binaries/"
+                    "{GITHUB_RELEASES}/jdk-{version}/{archive}"
+                    ).format(GITHUB_URL=GITHUB_URL, GITHUB_RELEASES=GITHUB_RELEASES,
+                    version=version, archive=self._archive)
+
+    def get_machine(self):
+        return mx.get_arch().replace("amd", "x") + "_" + mx.get_os().replace("darwin", "mac")
+
+class LabsJDKCE(JdkDistribution):
```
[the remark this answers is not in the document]
lazar-mitrovic: Fixed

mx.py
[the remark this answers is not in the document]
lazar-mitrovic: Fixed

mx_fetchjdk.py
lazar-mitrovic: Fixed

mx_fetchjdk.py
lazar-mitrovic: Fixed

mx_fetchjdk.py
lazar-mitrovic: Fixed

mx_fetchjdk.py
lazar-mitrovic: Fixed

mx_fetchjdk.py
lazar-mitrovic: Fixed

mx_fetchjdk.py
lazar-mitrovic: Fixed

mx_fetchjdk.py
lazar-mitrovic: Fixed

vjovanov:
Bump the mx version in mx.py please.

mx.py
lazar-mitrovic: Fixed

mx_fetchjdk.py
lazar-mitrovic: Fixed

mx_fetchjdk.py
lazar-mitrovic: Fixed

jdk_distribution_parser.py
lazar-mitrovic: Fixed

mx_fetchjdk.py
lazar-mitrovic: Fixed

mx_fetchjdk.py
lazar-mitrovic: Fixed

mx_fetchjdk.py
lazar-mitrovic: Fixed

mx_fetchjdk.py
lazar-mitrovic: Fixed

mx_fetchjdk.py
lazar-mitrovic: Fixed

mx_fetchjdk.py
lazar-mitrovic: Fixed

mx_fetchjdk.py
```diff
@@ -0,0 +1,204 @@
...
+
+    if not exists(final_path):
+        if not args["quiet"]:
+            print("Fetching {} archive from {}...".format(artifact, url))
+
+        archive_location = join(base_path, archive_name)
+        mx._opts.no_download_progress = args["quiet"]
+        sha1_hash = mx._hashFromUrl(sha_url).decode('utf-8')
+
+        if not exists(archive_location) or not args["keep-archive"]:
```
dougxc, line 65: So `--keep` silently does nothing when the JDK is already installed? It should print a warning:
```
The --keep option is ignored when the JDK is already installed.
```
If `--keep` is specified and the archive is available, its location should be printed.

vjovanov:
It does not show download progress anymore. Why?

mx_fetchjdk.py
```diff
@@ -0,0 +1,204 @@
...
+            shutil.rmtree(curr_path)
+            mx.abort("Error parsing archive. Please try again")
+
+        if not args["keep-archive"]:
+            os.remove(join(base_path, archive_name))
+            os.remove(archive_location + '.sha1')
+        os.rename(join(base_path, curr_path), final_path)
+
+    elif not args["quiet"]:
+        mx.warn("Requested JDK was already present")
```
dougxc, line 88:
```
        mx.log("Requested JDK is already installed at {}".format(final_path))
```

dougxc:
LGTM now.

PR #221, opened 2020-06-30
When not in a git repository look for .mx_vcs_root file

### LINKED ISSUE #217: mx requires primary suite to be in a VCS repository
## What happens:

mx fails to build suite when the suite is not in a VCS repository, which is often the case in CI/CD environments that use an archive of the source code instead of cloning the repository. 

## How to reproduce:
```bash
wget https://github.com/oracle/graal/archive/master.zip
unzip master.zip
cd graal-master/compiler
mx help
```

## Current Workaround
```diff
diff --git a/mx.py b/mx.py
index 8cf3518..75aa550 100755
--- a/mx.py
+++ b/mx.py
@@ -1083,7 +1083,7 @@ class SuiteImport:
         if version is None and version_from is None:
             if not (in_subdir and (importer.vc_dir != importer.dir or isinstance(importer, BinarySuite))):
                 abort("In import for '{}': No version given and not a 'subdir' suite of the same repository".format(name), context=context)
-            if importer.isSourceSuite():
+            if importer.isSourceSuite() and importer.vc_dir is not None:
                 suite_dir = join(importer.vc_dir, name)
             version = importer.version()
         if urls is None:
@@ -1490,8 +1490,6 @@ class Suite(object):
     :type dists: list[Distribution]
     """
     def __init__(self, mxDir, primary, internal, importing_suite, load, vc, vc_dir, dynamicallyImported=False):
-        if primary is True and vc_dir is None:
-            abort("The primary suite must be in a vcs repository")
         self.imported_by = [] if primary else [importing_suite]
         self.mxDir = mxDir
         self.dir = dirname(mxDir)
```

Relates to #216

zakkak:
When primary suite is not in a git repository look for the top most
directory containing a .mx_vcs_root file and use that directory as the
sources root.

Closes #217

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -911,10 +911,11 @@ class SuiteModel:
             _, primary_vc_root = VC.get_vc_root(suite_dir, abortOnError=False)
             if not primary_vc_root:
                 suite_parent = dirname(suite_dir)
-                # Use the heuristic of a 'ci.hocon' file being
+                # Use the heuristic of a 'ci.hocon' or '.mx_vcs_root' file being
                 # at the root of a repo that contains multiple suites.
                 hocon = join(suite_parent, 'ci.hocon')
-                if exists(hocon):
+                mx_vcs_root = join(suite_parent, '.mx_vcs_root')
+                if exists(hocon) or exists(mx_vcs_root):
                     return dirname(suite_parent)
                 return suite_parent
         else:
@@ -1518,7 +1519,7 @@ class Suite(object):
     """
     def __init__(self, mxDir, primary, internal, importing_suite, load, vc, vc_dir, dynamicallyImported=False):
         if primary is True and vc_dir is None:
-            abort("The primary suite must be in a vcs repository")
+            abort("The primary suite must be in a vcs repository or under a directory containing a file called '.mx_vcs_root' or 'ci.hocon'")
         self.imported_by = [] if primary else [importing_suite]
         self.mxDir = mxDir
         self.dir = dirname(mxDir)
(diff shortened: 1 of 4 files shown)


zakkak:
@gilles-duboscq I have pushed a first attempt of fixing cases where `mx` assumes a suite is under a vcs repository.
Could you please review it?

mx.py
```diff
@@ -10548,13 +10566,14 @@ def _addDevAttr(name, default=None):
...
                 l.warn("Library's suite version is too old to have maven metadata")
         pom.close('dependencies')
-    pom.open('scm')
-    scm = dist.suite.scm_metadata(abortOnError=validateMetadata != 'none')
-    pom.element('connection', data='scm:{}:{}'.format(dist.suite.vc.kind, scm.read))
-    if scm.read != scm.write or validateMetadata == 'full':
-        pom.element('developerConnection', data='scm:{}:{}'.format(dist.suite.vc.kind, scm.write))
-    pom.element('url', data=scm.url)
-    pom.close('scm')
+    if dist.suite.vc:
```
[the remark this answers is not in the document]
zakkak: Fixed in 10577

mx.py
```diff
@@ -14938,6 +14968,10 @@ def ignorePath(path, whitelist):
                 # skip whitelisted directories
                 dirnames[:] = []
                 continue
+            elif not suite.vc:
+                # skip suites not in a vcs repository
```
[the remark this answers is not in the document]
zakkak: Not sure, but it doesn't hurt :)

mx.py
```diff
@@ -1518,7 +1519,7 @@ class Suite(object):
     """
     def __init__(self, mxDir, primary, internal, importing_suite, load, vc, vc_dir, dynamicallyImported=False):
         if primary is True and vc_dir is None:
-            abort("The primary suite must be in a vcs repository")
+            abort("The primary suite must be in a vcs repository or under a directory containing a file called '.mx_vcs_root'")
```
zakkak, line 1522: @gilles-duboscq I also added this to reflect the `.mx_vcs_root` file support

zakkak:
`master` seems to not pass travis CI.

gilles-duboscq:
I think this looks good. I will take it for further testing.
It will not solve the issues from other suites such as `sdk`, `substratevm`, `compiler`, and `graalpython` but we have to start at the `mx` level.

One thing though: just like OpenJDK, I don't think we want to add a copyright line for each contributor outside of the case where the whole file comes from a contribution.

zakkak:
> I think this looks good. I will take it for further testing.

Thanks @gilles-duboscq !

> It will not solve the issues from other suites such as sdk, substratevm, compiler, and graalpython but we have to start at the mx level.

Correct, I volunteer to work on them as well after this gets merged.

> One thing though: just like OpenJDK, I don't think we want to add a copyright line for each contributor outside of the case where the whole file comes from a contribution.

Noted. I removed the copyright changes.

gilles-duboscq:
Thank you.
This looks good and passed internal testing, I will schedule it for integration.

mx.py
```diff
@@ -911,10 +911,10 @@ def siblings_dir(suite_dir):
             _, primary_vc_root = VC.get_vc_root(suite_dir, abortOnError=False)
             if not primary_vc_root:
                 suite_parent = dirname(suite_dir)
-                # Use the heuristic of a 'ci.hocon' file being
+                # Use the heuristic of a '.mx_vcs_root' file being
                 # at the root of a repo that contains multiple suites.
-                hocon = join(suite_parent, 'ci.hocon')
-                if exists(hocon):
+                mx_vcs_root = join(suite_parent, '.mx_vcs_root')
```
gilles-duboscq, line 916: during internal review, it came up that we should rather check for both files (`ci.hocon` or `.mx_vcs_root`) to avoid any breaking change to thing that might have relied on the existing `ci.hocon` heuristic.

zakkak: Thanks @gilles-duboscq !
This should now be resolved.

PR #222, opened 2020-07-02
Fix --tags ranges regex

zakkak:


dougxc:
Looks good - merging now.

dougxc:
@zapster pointed out that the current implementation matches what the documentation says:
```
  --tags TAGS           comma separated list of tags to select subset of tasks
                        to be run. Tags can have a range specifier
                        `name[:from:[to]]`.If present only the [from,to) tasks
                        are executed. If `to` is omitted all tasks starting
                        with `from` are executed.
```

That is, `--tags:style:4:` is the way to specify "all style tasks starting at the 4th".

zakkak:
I might need to increase my font size (or retrain my brain) :D

Thanks @zapster for catching this.

zapster:
@zakkak I also had to look at the regex and the --help message a few times before noticing it. :)

PR #229, opened 2020-10-30
Ensure select_jdk.sh finds the env file when --suite-dir is used

a74nh:
The existing code always used . when finding suite.py


dougxc:
Thanks for the fix. I'll integrate it.

PR #230, opened 2020-12-09
Linux-musl support

petermz:
This patch adds support for musl-based Linux distros such as Alpine. This is needed in order to enable musl-based GraalVM builds.

I've chosen rather simple approach and just introduced a new `os_arch`: `linux-musl`. It is detected by parsing `ldd` output. Because it's more costly than just examining `sys.platform`, the result is cached in a global var. This new arch type is used right away in `suite.py` to define musl-based Ninja binary.

This solution is mostly backward compatible, I don't expect it to break many existing projects. Builds may start failing when run on musl-based Linux (because the `linux-musl` key is lacking), but they would have failed anyway later when they try to execute glibc-based binaries they'd have downloaded from the `linux` section.

### DIFF
--- a/mx.mx/suite.py
+++ b/mx.mx/suite.py
@@ -519,6 +519,12 @@ suite = {
             "optional" : True
           }
         },
+        "linux-musl" : {
+          "amd64" : {
+            "urls" : ["https://download.bell-sw.com/graalvm/ninja-1.10.1-x86_64-linux-musl.zip"],
+            "sha1" : "0f3d92566d116edc06e7a983ec1bbc4d4a9e9c2d"
+          }
+        },
         "darwin" : {
           "amd64" : {
             "urls" : ["https://github.com/ninja-build/ninja/releases/download/v{version}/ninja-mac.zip"],
--- a/mx.py
+++ b/mx.py
@@ -3853,7 +3853,7 @@ def get_os():
     Get a canonical form of sys.platform.
     """
     global _os
-    if not _os:
+    if _os is None:
         _os = _get_os()
         if _opts and _opts.verbose:
             log('OS detected: %s' % _os)
--- a/mx.py
+++ b/mx.py
@@ -2176,9 +2176,11 @@ class Suite(object):
     def _pop_os_arch(attrs, context):
         os_arch = attrs.pop('os_arch', None)
         if os_arch:
-            os_key = get_os_variant()
-            if not os_key in os_arch and '-' in os_key:
-                os_key = os_key.split('-')[0]
+            os_key = None
+            if get_os_variant():
+                os_key = get_os() + '-' + get_os_variant()
+            if os_key is None or not os_key in os_arch:
+                os_key = get_os()
             if not os_key in os_arch:
                 os_key = '<others>'
             os_attrs = os_arch.pop(os_key, None)
@@ -2191,7 +2193,7 @@ class Suite(object):
                 else:
                     warn("No platform-specific definition is available for {} for your architecture ({})".format(context, get_arch()))
             else:
-                warn("No platform-specific definition is available for {} for your OS ({})".format(context, get_os_variant()))
+                warn("No platform-specific definition is available for {} for your OS ({})".format(context, get_os()))
         return None
 
     @staticmethod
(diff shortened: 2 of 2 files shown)


petermz:
I may have signed the OCA as peter.zhelezniakov@oracle.com

mx.mx/suite.py
```diff
@@ -519,6 +519,12 @@
             "optional" : True
           }
         },
+        "linux-musl" : {
+          "amd64" : {
+            "urls" : ["https://download.bell-sw.com/graalvm/ninja-1.10.1-x86_64-linux-musl.zip"],
```
zapster, line 524: We would need to rebuild/repackage this on our own infrastructure. Is there anything special about this? Could you share your build instructions?

zapster:
Just to be sure, this means that every dependency with `os_arch/linux` will need to get an `os_arch/linux_musl` entry, right? That is probably ok in most cases since almost everything depends on libc. However, there might be cases where this is not the true. Maybe using `linux` in case where there is no `linux_musl` would be more appropriate? Of course, this adds more complexity.

mx.mx/suite.py
petermz: I believe I've just cloned their repo and run `./configure.py --bootstrap`

petermz:
> Just to be sure, this means that every dependency with `os_arch/linux` will need to get an `os_arch/linux_musl` entry, right?

If they plan to build on linux-musl, yes

> Maybe using `linux` in case where there is no `linux_musl` would be more appropriate? Of course, this adds more complexity.

Surely we can fallback to `linux`, the question is, is that useful? It would be for some dependency that is platform dependent (otherwise there's no reason to have `os_arch` at all), but not libc dependent. So, maybe some driver firmware or a shell script. OTOH a fallback could mask errors in the project, e.g. when some binary dependency is redistributed

petermz:
I've realized that `get_os()` is used for things other than binary dependencies, such as paths and compiler flags. So falling back from `linux-musl` to `linux` seems indeed useful. I've added it in the latest commit.

gilles-duboscq:
Thanks for this proposal @petermz.

* I think we should indeed fallback. Hopefully most native projects can be built with no libc-specific options and native libraries are not that common. I think in this regard we should aim to make all libc specific options mostly disappear from projects by having good enough defaults in the native project support. Libraries are the only thing that really require libc-specific data in `suite.py` files.
* Can we make `get_os_variant` only return the variant part?

We have to be careful with such changes since mx is always backward compatible to this means we'll have to support this model for a long time. Also we should be careful about the fact that we will likely want to allow cross-compilation.

petermz:
Fixed `get_os_variant`, and in the commit just before that, I've added the fallback

mx.py
```diff
@@ -3854,11 +3856,12 @@ def get_os():
...
     global _os_variant
     if _os_variant is None:
-        _os_variant = get_os()
-        if _os_variant == 'linux' and 'musl' in _check_output_str(['ldd', sys.executable]):
-            _os_variant = 'linux-musl'
+        if get_os() == 'linux' and 'musl' in _check_output_str(['ldd', sys.executable]):
+            _os_variant = 'musl'
+        else:
+            _os_variant = ''
         if _opts and _opts.verbose:
```
gilles-duboscq, line 3863: Any reason why you don't use `logv` instead of checking for `_opts.verbose`?

mx.py
```diff
@@ -3854,11 +3856,12 @@ def get_os():
...
-            log('OS detected: %s' % _os_variant)
+            log('OS variant detected: %s' % (_os_variant if _os_variant else 'none'))
```
gilles-duboscq, line 3864: We usually favour `str.format`

gilles-duboscq (commented):
Beside some minor comments i think this should work.

mx.py
petermz: No, just reused this pattern from somewhere I saw it

mx.py
```diff
@@ -2176,9 +2176,14 @@ def _pop_list(attrs, name, context):
     def _pop_os_arch(attrs, context):
         os_arch = attrs.pop('os_arch', None)
         if os_arch:
-            os_attrs = os_arch.pop(get_os(), None)
-            if not os_attrs:
-                os_attrs = os_arch.pop('<others>', None)
+            os_key = None
+            if get_os_variant():
+                os_key = get_os() + '-' + get_os_variant()
+            if os_key is None or not os_key in os_arch:
```
gilles-duboscq, line 2182: Sorry for not noticing this in my previous pass: some linters like pep8 will complain about `not x in y` asking to write it as `x not in y` (same below).

After that change i think this is ready for integration (we can also do that style change during integration on our side).

petermz:
Thanks Gilles for your review. I've fixed both things in the latest commit.
Regarding OCA, as I said above, I believe I'm covered with Bellsoft blanket agreement, so hopefully this check can be cleared

petermz:
No problem, I've made all the changes. Thanks Gilles for your review!
Regarding OCA, as I said above, I believe I'm covered with Bellsoft blanket agreement, so hopefully this check can be cleared

petermz:
Is there anything left to do here?
I thought about bumping mx version maybe? So that I can enforce it in other places

gilles-duboscq:
This was merged in c41f3403a10bd9e89e5132924513f5cddf66d2f3, the version was bumped to 5.284.0.
Sorry this took longer than expected.

PR #233, opened 2021-04-10
Use explicit encoding for reading from file.

FilipJirsak:
open() without encoding uses platform default encoding. On WIndows it can be encoding which has some bytes undefined - for example Windows-1250 don't know 0x88.

### DIFF
--- a/mx_ide_eclipse.py
+++ b/mx_ide_eclipse.py
@@ -130,7 +130,7 @@ def eclipseformat(args):
     class FileInfo:
         def __init__(self, path):
             self.path = path
-            with open(path) as fp:
+            with open(path, encoding='utf-8') as fp:
                 self.content = fp.read()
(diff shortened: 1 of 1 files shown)


mx_ide_eclipse.py
```diff
@@ -130,7 +130,7 @@ def read_core_prefs_file(self):
     class FileInfo:
         def __init__(self, path):
             self.path = path
-            with open(path) as fp:
+            with open(path, encoding='utf-8') as fp:
```
dougxc, line 133: This does not work on python2:
```
  File "/b/b/e/path with a space/mx_ide_eclipse.py", line 133, in __init__
    with open(path, encoding='utf-8') as fp:
TypeError: 'encoding' is an invalid keyword argument for this function
```
We are working to make mx require python 3.6 but are not quite there yet.

PR #235, opened 2021-05-15
Fix typo in README

miguelaferreira:
This PR removes a rogue ` (back-tick) char from the README.


miguelaferreira:
I've signed the agreement

PR #241, opened 2021-08-16
[GR-33052] Allow get_runtime_jvm_args to exclude implicit jars.

graalvmbot:


mx.py
```diff
@@ -12066,12 +12066,16 @@ def classpath(names=None, resolve=True, includeSelf=True, includeBootClasspath=F
     return _entries_to_classpath(cpEntries=cpEntries, resolve=resolve, includeBootClasspath=includeBootClasspath, jdk=jdk, unique=unique, ignoreStripped=ignoreStripped)
 
 
-def get_runtime_jvm_args(names=None, cp_prefix=None, cp_suffix=None, jdk=None):
+def get_runtime_jvm_args(names=None, cp_prefix=None, cp_suffix=None, jdk=None, exclude_names=None):
```
dougxc, line 12069: Why can't `names` be filtered by the caller to exclude `exclude_names`? If there's good reason for this otherwise redundant parameter, then it should be documented.

PR #244, opened 2021-09-29
Support for overriding SHA1 with URL rewriting

ceresek:
When overriding a resource URL with `urlrewrites` or `MX_URLREWRITES`, it is not possible to substitute a modified resource with a different SHA1 because the subsequent SHA1 check fails. This patch extends the URL rewriting functionality with the possibility to specify a new SHA1 value to be checked against. It is useful for example when running modified benchmarks from local JAR files.

Possible issue. The URL rewriting rules are associated with URL patterns, SHA1 is associated with resources, a resource can have multiple URL values specified. This is probably not a problem from usability perspective, but the code modifications are somewhat awkward because URL rewriting and SHA1 checks happen at slightly different levels.

Tested only with native image benchmarking commands, may need testing with other suites.


ceresek:
Sorry, my bad. I've signed the OCA with `petr.tuma@acm.org`. Should I resubmit the PR with a new mail ?

ceresek:
I've submitted the OCA version 1.7.1 in November 2019 for petr.tuma@acm.org and it worked back then.

dougxc:
Hi Petr, thanks for the PR. Could you please also update the "URL rewriting" section in README.md.

dougxc:
> Possible issue. The URL rewriting rules are associated with URL patterns, SHA1 is associated with resources, a resource can have multiple URL values specified. This is probably not a problem from usability perspective, but the code modifications are somewhat awkward because URL rewriting and SHA1 checks happen at slightly different levels.

What exactly is the problem? I assume one will just need to ensure there's a rewrite rule for each URL of a resource whose sha1 you want to modify?

ceresek:
> Hi Petr, thanks for the PR. Could you please also update the "URL rewriting" section in README.md.

Hi Doug, the doc was already updated, let me know if more info is needed ?

ceresek:
> What exactly is the problem? I assume one will just need to ensure there's a rewrite rule for each URL of a resource whose sha1 you want to modify?

Actually, specifying SHA1 with the first URL should be enough. Since the SHA1 check is in a different place than the URL processing loop, the SHA1 rewriting code does not "know" what URL will be used. But in essence yes, it is not a usability problem per se.

dougxc:
> Hi Doug, the doc was already updated, let me know if more info is needed ?

Sorry, I missed that. I thought I was looking at code comments.
I'll merge this PR.

Issue #232, opened 2021-03-13
When to support ARM 64?


jimmysuncpt:
When I run `mx` on my MacBook M1, it warns that "unknown or unsupported architecture: os=darwin, machine=arm64".
So, is there any plan to support ARM 64?

dougxc:
Eventually mx will support this architecture. Currently, no one on the GraalVM team (that I'm aware of) has such a machine. Maybe you want to submit a PR?
If you are asking about GraalVM support for M1 (as opposed to mx support), then that will be quite some more effort but it's also something we plan to support as resources become available.

fniephaus:
Feel free to follow https://github.com/oracle/graal/issues/2666 for updates on M1 support.

PR #245, opened 2021-10-01
Add --arch global option to support arm64 macOS via Rosetta2

errikos:
Right now, `mx` refuses to run on M1 Macs (#232).

Allow to specify a custom architecture target. Useful for arm64-based Apple computers, until all used native libraries are released for arm64.

Example usage:
  mx --arch amd64 ...

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -581,6 +581,7 @@ environment variables:
                                 "projects and store it in the given <file>. If <file> is 'default', the compilation database will "
                                 "be stored in the parent directory of the repository containing the primary suite. This option "
                                 "can also be configured using the MX_COMPDB environment variable. Use --compdb none to disable.")
+        self.add_argument('--arch', action='store', dest='arch', help='force use of the specified architecture')
 
         if not is_windows():
             # Time outs are (currently) implemented with Unix specific functionality
@@ -678,6 +679,10 @@ environment variables:
                         pass
                 except IOError as e:
                     abort('Error opening {} specified by --exec-log: {}'.format(opts.exec_log, e))
+
+            system_arch = platform.uname()[4]
+            if opts.arch and opts.arch != system_arch:
+                warn('overriding detected architecture ({}) with {}'.format(system_arch, opts.arch))
 
         else:
             parser = ArgParser(parents=[self])
@@ -3835,6 +3840,9 @@ def _separatedCygpathW2U(p):
     return os.pathsep.join(map(_cygpathW2U, p.split(';')))
 
 def get_arch():
+    return _opts.arch if _opts.arch else _get_real_arch()
+
+def _get_real_arch():
     machine = platform.uname()[4]
     if machine in ['aarch64']:
         return 'aarch64'


errikos:
> Hello Ergys Dona, thanks for contributing a PR to our project!
> 
> We use the [Oracle Contributor Agreement](http://www.oracle.com/technetwork/community/oca-486395.html) to make the copyright of contributions clear. We don't have a record of you having signed this yet, based on your email address ergys -(dot)- dona -(at)- gmail -(dot)- com. You can sign it at that link.
> 
> If you think you've already signed it, please comment below and we'll check.

One of my organisations here at GitHub (HexHive at EPFL, my employer) has signed the CLA.

errikos:
As already state above, one of my organisations here at GitHub (HexHive at EPFL - my employer, https://github.com/HexHive) has signed the CLA, so I have been told that it should not be necessary that I sign it individually.

dougxc:
This change is too intrusive and risky to be made as an automatic feature. How about adding an `--arch` global option to mx to force the arch value selected?

errikos:
* Added a global `--arch` option.
* Introduced a new function called `get_effective_arch()` to get the architecture that will be effectively used in operation. This is to keep the `get_arch()` semantics unmodified.
* Replaced the `get_arch()` calls with `get_effective_arch()`.

There are currently no restrictions in the architecture that can be specified, except from the list of allowed options.

dougxc:
We don't want to introduce a new concept of "effective arch" and now all callers now need to think about the diff between `get_arch` and `get_effective_arch`. Can you think of a scenario under which 2 different answers make sense in a single mx execution?

errikos:
> We don't want to introduce a new concept of "effective arch" and now all callers now need to think about the diff between `get_arch` and `get_effective_arch`. Can you think of a scenario under which 2 different answers make sense in a single mx execution?

To be honest the only reason for this change is to display the real underlying architecture in the introduced warning.

We could instead just introduce an "internal" method `_get_real_arch` and leave the callers as they were.

dougxc:
That looks much better.
@gilles-duboscq does this look good to you?

mx.py
```diff
@@ -581,6 +581,7 @@ def __init__(self, parents=None):
                                 "projects and store it in the given <file>. If <file> is 'default', the compilation database will "
                                 "be stored in the parent directory of the repository containing the primary suite. This option "
                                 "can also be configured using the MX_COMPDB environment variable. Use --compdb none to disable.")
+        self.add_argument('--arch', action='store', dest='arch', help='force use the specified architecture', choices=['arm64', 'aarch64', 'amd64', 'sparcv9'])
```
gilles-duboscq, line 584: Why allow `arm64` in there? we usually canonicalize to `aarch64`

errikos: Agreed, it's just because Apple calls their processors "arm64" (IIRC). We could remove it if we don't like to have an alias (as far as I know they are the same).

dougxc: Let's keep mx arches canonical until there's a good reason not to.

gilles-duboscq:
LGTM

mx.py
```diff
@@ -581,6 +581,7 @@ def __init__(self, parents=None):
                                 "projects and store it in the given <file>. If <file> is 'default', the compilation database will "
                                 "be stored in the parent directory of the repository containing the primary suite. This option "
                                 "can also be configured using the MX_COMPDB environment variable. Use --compdb none to disable.")
+        self.add_argument('--arch', action='store', dest='arch', help='force use the specified architecture', choices=['aarch64', 'amd64', 'sparcv9'])
```
teshull, line 584: Why allow sparcv9? Do we still need that for some compatibility?

teshull (changes requested):

mx.py
errikos: `get_arch` could return that, so that's why I put it in the allowed choices.

.metals/metals.lock.db
```diff
@@ -0,0 +1,6 @@
+#FileLock
```
dougxc, line 1: Was this file included by mistake?

errikos: Indeed.

PR #248, opened 2022-01-05
Add detection of `python3` binary.

nirvdrum:
Python 2 is being or has been phased out of default installations on many systems. [PEP 394](https://legacy.python.org/dev/peps/pep-0394/) means systems such as Ubuntu will not alias `python3` to `python` by default. In order to use the Python 3 binary with mx, the end user must either set `MX_PYTHON=python3`, `MX_PYTHON_VERSION=3`, or set up an alias by some other means. It would be helpful if mx checked if `python3` exists and use it, as part of its existing executable search.

This should maintain full backwards-compatibility. The `python3` binary is only chosen if `python2.7`, `python2`, and `python` do not exist.

### DIFF
--- a/mx
+++ b/mx
@@ -60,7 +60,18 @@ else
         if [ $? -eq 0 ]; then
             python_exe=python2
         else
-            python_exe=python
+            type python > /dev/null 2>&1
+            if [ $? -eq 0 ]; then
+                python_exe=python
+            else
+                type python3 > /dev/null 2>&1
+                if [ $? -eq 0 ]; then
+                    python_exe=python3
+                else
+                    echo "Cannot find Python on PATH"
+                    exit 1
+                fi
+            fi
         fi
     fi
 fi


zakkak:
I think python3 should be preferred when available even if that breaks "full backwards-compatibility" as python2 has reached its end of life.

nirvdrum:
@dougxc @gilles-duboscq Can we favor Python 3 over 2? If so, I'll reorder the executable checks in this PR.

Additionally, that Python 3.x gate failure looks unrelated to this PR.

fniephaus:
> Can we favor Python 3 over 2?

IIRC we looked into exactly this only recently and ran into a bunch of problems that need to be addressed first (or simultaneously). Anyway, I agree that it needs to happen sooner rather than later.

fniephaus:
Assigning to @dougxc who is currently working on this. We may be dropping Python 2 support entirely soon.

nirvdrum:
Thanks. Regardless of dropping Python 2 support, this PR would help environments where only Python 3 is installed.

dougxc:
I agree that merging this in the interim is a good idea. Would you be able to take care of that @fniephaus while I focus on the larger change to drop python2 altogether?

BTW, I've opened #249 to track the larger change.

PR #263, opened 2022-06-06
Build SVM with JDK 19.

oraluben:
Ref: oracle/graal#4623

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -7670,6 +7670,8 @@ class JavacLikeCompiler(JavaCompiler):
             javacArgs += ['-proc:none']
         c = str(compliance)
         javacArgs += ['-target', c, '-source', c]
+        if c == '19' and self.jdk.javaCompliance == '19':
+            javacArgs += ['--enable-preview']
         if _opts.very_verbose:
             javacArgs.append('-verbose')
 


oraluben:
Hi @dougxc , the current version of `requiresConcealed` seems to have issue.
IIUC it's checking if the current `jdk.javaCompliance` is in the range of `java_compliance`, i.e., 
```
if java_compliance.__contains__(jdk.javaCompliance)
```
, i.e.,
```
if jdk.javaCompliance in java_compliance
```
. Instead mx now has (negation is not important here):

https://github.com/graalvm/mx/blob/c900245b403df797a8e9fa83597cfeb97482a0a7/mx_javamodules.py#L1121

I met this issue while having something like:
https://github.com/oracle/graal/blob/e0f7b9735df345077373cdf82df499e923ca30d3/substratevm/mx.substratevm/suite.py#L229-L236


Does this fix look good to you: https://github.com/graalvm/mx/pull/263/commits/e4128a5bf56273c1bcf45fb8c58b0fc3844f0502

mx.py
```diff
@@ -7670,6 +7670,8 @@ def prepare(self, sourceFiles, project, outputDir, classPath, processorPath, sou
             javacArgs += ['-proc:none']
         c = str(compliance)
         javacArgs += ['-target', c, '-source', c]
+        if c == '19' and self.jdk.javaCompliance == '19':
```
dougxc, line 7673: I think this should be `>=` instead of `==`. What happens when we get to Java 20 and loom is still in preview?

mx_javacompliance.py
```diff
@@ -213,9 +206,19 @@ def __cmp__(self, other):
...
+        :return: if `other` is satisfied by `self`
+        """
         if isinstance(other, (int, str)):
             other = JavaCompliance(other)
-        assert other._high_bound() is not None, "Contains check cannot be done with version ranges"
+        assert (
+                len(other._parts) == 1 and
+                (
+                        other._high_bound() is None or
+                        other._high_bound() == other.value
```
dougxc, line 219: How can this happen? Can you please add extra tests to `_test()` in `mx_javacompliance.py` to cover this case as well as any other cases made relevant by your changes.

mx.py
oraluben: This PR is not ready for review, and this is a temporary solution for local tests. Besides, I think the preview feature in JDK is very tricky for Graal, E.g. `jdk.internal.access.foreign.MemorySegmentProxy` has been moved to a preview feature in JDK-8282191, which affects some substitutions in `Target_java_nio_DirectByteBuffer_JDK17OrLater.java`.

mx_javacompliance.py
oraluben: Sure, I'll add tests for range checks.

PR #264, opened 2022-06-27
Refine range check in JavaCompliance.

oraluben:
This PR updates `JavaCompliance`, including:
1. Better `JavaCompliance.__contains__`. It now checks every parts of `JavaCompliance` instances, rather than only the high and low bounds. `mx.JavaCompliance("16") in mx.JavaCompliance("11..15,17")` now returns `False` as expected;
2. Fix incorrect version check when having `requiresConcealed` with java compliance specified, this also helps to port Graal to JDK 19.

### DIFF
--- a/mx_javacompliance.py
+++ b/mx_javacompliance.py
@@ -104,6 +104,15 @@ class JavaCompliance(mx.Comparable):
                 if self._high is None:
                     return True
                 return value <= self._high
+            elif isinstance(other, JavaCompliance._Range):
+                if other._high is None:
+                    # open range is considered not match, e.g. '1.8+' not in '1.8..11'
+                    return False
+                if other._low < self._low:
+                    return False
+                if self._high is None:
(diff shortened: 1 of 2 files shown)


oraluben:
@dougxc This is the ready-to-review part of #263, with range tests included.

dougxc:
@gilles-duboscq could you please help review this as well.

oraluben:
> You may even remember writing them ;-)

Now I do 🤣

Anyways, the Loom cleanup is not so urgent at the moment, but I do hope the java compliance changes to be merged if you think it's ok, I'll remove the loom changes from this PR.

gilles-duboscq:
I agree we need to keep the loom flags and behaviors, the other changes look good.

dougxc:
Resolved by 49f40cad895bc2603407f15057b66d0d6d8e1baa.

PR #271, opened 2022-11-07
Fix URL rewrites with support for more digests

ceresek:
The changes in 74fd67ab233d5a7447215213944f96204fc2bf64 left a cast in from when `digest` was just a string, this breaks when `digest` is an instance of `Digest`. Looks like simple removal of the cast works fine (at least it does in our local infrastructure).

### DIFF
--- a/mx_urlrewrites.py
+++ b/mx_urlrewrites.py
@@ -192,7 +192,7 @@ class URLRewrite(object):
         # Make sure to use str rather than unicode.
         # Some code paths elsewhere depend on this.
         self.replacement = str(replacement)
-        self.digest = str(digest) if digest else None
+        self.digest = digest if digest else None
 
     def _rewrite(self, url):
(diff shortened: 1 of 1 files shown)


ceresek:
I have signed the OCA under the commit mail <petr.tuma@acm.org>. Since this is a trivial update, maybe doing it on your side would be more flexible than me going through the signing process again ?

dougxc:
> @dougxc I have signed the OCA under the commit mail [petr.tuma@acm.org](mailto:petr.tuma@acm.org). Since this is a trivial update, maybe doing it on your side would be more flexible than me going through the signing process again ?

Will do. I'll slightly change the fix to just be `self.digest = digest` as well.

ceresek:
Thank you :-)

>  I'll slightly change the fix to just be self.digest = digest as well.

You compiler people are optimizing everywhere :-)

PR #273, opened 2023-02-03
Support for musl-based Linux

petermz:
A couple of fixes related to musl-based Linux support:
* Changed the order in which OS attributes are examined: most specific variant should come first, e.g. `linux-musl`, then `linux`, then `<others>`
* Added `-Wno-error=cpp` for `linux-musl` so that it doesn't complain about nonstandard headers such as `<sys/errno.h>`

This is needed to take advantage of https://github.com/oracle/graal/pull/3141

### DIFF
--- a/java/com.oracle.jvmtiasmagent/src/jvmtiasmgent.c
+++ b/java/com.oracle.jvmtiasmagent/src/jvmtiasmgent.c
@@ -29,7 +29,7 @@
 #include <time.h>
 #include <limits.h>
 #include <pthread.h>
-#include <sys/errno.h>
+#include <errno.h>
 #include <arpa/inet.h>
 
 #include <jni.h>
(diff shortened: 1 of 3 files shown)


gilles-duboscq:
Wouldn't it be better to include `<errno.h>`?
/cc @tkrodriguez

petermz:
They should be interchangeable on Linux (same for poll.h, signal.h etc), but I'm not sure about Windows and Darwin

gilles-duboscq:
Looks like that should be fine for [Windows](https://learn.microsoft.com/en-us/cpp/c-runtime-library/errno-constants?view=msvc-170) and a quick check on a macOS machine shows that it has `/usr/include/errno.h` (and `/Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk/usr/include/errno.h`)

gilles-duboscq (approved):

gilles-duboscq:
I'll start integrating that. I'll take care of bumping the mx version number.

gilles-duboscq:
You can now depend on the updated behaviour by requesting a minimal mx version of 6.15.0

PR #277, opened 2023-08-01
Add --update-suite  to sforceimport to allow updates without user interaction

smarr:
I'd like to be able to use `sforceimport` to update my `suite.py` in a cronjob, without user interaction.

I see two options:
 1. add a new flag like this PR, `--update-suite` or similar to sforceimport
 2. or add a default value for `ask_yes_no`, though, this option changes the current behavior minimally, which might be more risky. Don't know.

This is useful for scripts since ask_yes_no checks for an interactive session and bails out without one, if there's no default value.

### DIFF
--- a/mx.py
+++ b/mx.py
@@ -17043,11 +17043,11 @@ def sbookmarkimports(args):
         primary_suite().visit_imports(_sbookmark_visitor)
 
 
-def _scheck_imports_visitor(s, suite_import, bookmark_imports, ignore_uncommitted, warn_only):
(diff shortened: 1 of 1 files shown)


smarr:
Hm, the gate is failing. It looks like `jq` is trying to find an pip key in `common.json`, which has been failing for a while.

eregon:
Isn't mx's option `  -y                            answer 'y' to all questions asked` enough?

smarr:
Yepp, I overlooked that. Thanks!

PR #279, opened 2024-01-08
[GR-51173] Improve computation of suite root dirs.

ansalond:
To locate the root dir of a suite, mx:
* checks if the suite is contained in one or more multiple VCS, returning the one with the deepest nesting
* if no VCS is found, looks at .mx_vcs_root and ci.hocon files

We should try both approaches and return the "most nested" result (that is, the longest path, which we assume it's the closest to the suite).

### DIFF
--- a/common.json
+++ b/common.json
@@ -4,7 +4,7 @@
     "Jsonnet files should not include this file directly but use ci/common.jsonnet instead."
   ],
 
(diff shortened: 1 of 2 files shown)


zakkak (approved):
LGTM, thanks @ansalond

PR #280, opened 2024-01-30
Make note about MX_BUILD_EXPLODED conditional on it not being true

smarr:
This is really a very minor things, though, every time I do a `mx eclipseinit` and it reminds me of MX_BUILD_EXPLODED, I spend a few minutes figuring out whether I got the config wrong.

By adding the check, the note is only printed when the configuration is something other than `MX_BUILD_EXPLODED=true`.


smarr:
Hm, actually, I can't make anything work with `MX_BUILD_EXPLODED=true` set. Neither my own language nor GraalPy. All break with different issues. Some Java module issue or some dynamic library issue on GraalPy.

Issue #293, opened 2025-08-01
Error while doing maven-install


javaman-dev:
line 18163, in maven_install

  mvn_local_install(dist.maven_group_id(), dist.maven_artifact_id(), dist.path, releaseVersion, args.repo)
                                                                                                                      ^^^^^^^^^
AttributeError: 'POMDistribution' object has no attribute 'path'

PR #294, opened 2025-09-03
Fix for "Error while doing maven-install "

JohnTortugo:
I hit the error described on #293 while doing some local tests, the fix for me was simply to skip the `POMDistribution` entries. I'm no MX expert but this seems a reasonable fix. I tested it on OSX/Linux AArch/x86.

Closes: #293

### DIFF
--- a/src/mx/_impl/mx.py
+++ b/src/mx/_impl/mx.py
@@ -18206,7 +18206,7 @@ def maven_install(args):
         dists = [d for d in s.dists if _dist_matcher(d, None, False, only, None, False)]
         for dist in dists:
             # ignore non-exported dists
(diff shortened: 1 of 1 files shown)


dougxc:
@gilles-duboscq would you mind looking at this.

gilles-duboscq:
I see. It would be helpful to check what exactly is triggering this internet connection.
I initially thought that the difference was in the maven command used but i see that `mx maven-deploy` already uses `install:install-file` (same as what `mx maven-install` uses) instead of `deploy:deploy-file` or `gpg:sign-and-deploy-file` when deploying to the local repository.

Are you explicitly specifying the path to the local repo (`mx maven-install --repo /path/to/repo` or `mx maven-deploy local file:///path/to/repo`)? I think the special "local" code path is not taken in that case in `mx maven-deploy`. If that's what's causing your issue we could fix that.

PR #298, opened 2026-02-03
Add support for custom Maven repository URLs

zakkak:
In some build environments it's required to use custom maven repositories (e.g., internal mirrors). This PR enables the global override of the default maven repositories to satisfy that requirement.

cc @jerboaa


jerboaa:
Any update on this? It would be nice to have a way to specify additional maven repos via an environment variable rather than have it hard-coded.

jerboaa:
cc @dougxc 

Do you think something like this would be acceptable?

dougxc:
I don't see a problem with it since maven dependencies used by mx all have a digest.

@ezzarghili can you see any other risks?

PR #305, opened 2026-06-03
Add support for patching JDK modules

thomaswue:
Summary:
- Add `--patch-jdk-module <module>=<path>` for patching JDK modules from mx.
- Apply configured patches to Java launches, JDK module discovery, javac-like compilation, and generated `module-info.java` compilation.
- Include patch path metadata in the cached JDK module list source so patched module discovery is refreshed when the patch changes.

Review plan:
1. Start with option parsing and patch helper functions in `src/mx/_impl/mx.py`.
2. Review the JDK module discovery/cache and javac-like compilation changes in `src/mx/_impl/mx.py`.
3. Finish with generated module-info compilation in `src/mx/_impl/mx_javamodules.py`.

Local checks:
- Built a scratch project importing a package added through `--patch-jdk-module`.
- Verified `mx java --describe-module` observes an added patch package.
- Ran `python3 -m py_compile` on the touched Python files and a focused mx build of `com.oracle.mxtool.compilerserver`.

### DIFF
--- a/src/mx/_impl/mx.py
+++ b/src/mx/_impl/mx.py
@@ -904,6 +904,8 @@ environment variables:
         self.add_argument('--jmods-dir', action='store', help="path to directory containing jmods which are added to --module-path when compiling "
                                 "module-info.java for a distribution's module. The default is JAVA_HOME/jmods. Specify NO_JMODS to indicate "
                                 "JAVA_HOME is a JEP 493-enabled JDK (i.e. has no jmods).", metavar='<path>')
+        self.add_argument('--patch-jdk-module', action='append', dest='jdk_module_patches', metavar='<module>=<path>', default=[],
+                          help='patch a JDK module for mx Java launches, JDK module discovery, and Java source compilation')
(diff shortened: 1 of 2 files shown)


src/mx/_impl/mx.py
```diff
@@ -11784,6 +11789,62 @@ def extract_VM_args(args, useDoubleDash=False, allowClasspath=False, defaultAllV
...
+def _jdk_module_patch_args(patches):
+    patches = tuple(patches)
+    if not patches:
+        return []
+    return ['--add-modules=' + ','.join(module for module, _ in patches)] + [
+        '--patch-module=' + module + '=' + patch_path for module, patch_path in patches
+    ]
+
+
+def _jdk_module_patch_signature(patches):
```
dougxc, line 11825: This is computing a fingerprint for a cache key so a better name would be `_jdk_module_patch_cache_key`

dougxc:
It would be useful to add the motivation/background for this feature to the PR description.

dougxc (approved):

zakkak:
Hi @dougxc the context behind this PR is in https://github.com/graalvm/mandrel/discussions/979. I suggest not merging this PR though as this is only a small part of the changes needed to make it work and we are still in the exploration phase.

PR #308, opened 2026-07-15
Add ECJ version compatible with Java 25

eregon:
* ECJ 3.43.0 with `"javaCompliance": "25+"` fails with: target level should be in '1.8','9'...'24' (or '8.0'..'24.0'): 25
* ECJ 3.44.0 emits incorrect warnings and errors.
* ECJ 3.45.0 and 3.46.0 work fine, I picked 3.46.0 (the latest release).

### DIFF
--- a/mx.mx/suite.py
+++ b/mx.mx/suite.py
@@ -1097,13 +1097,24 @@ suite = {
       "licence": "EPL-2.0",
     },
 
-    # compatible version for JDK 24..25
+    # compatible version for JDK 24
     "ECJ_3.43": {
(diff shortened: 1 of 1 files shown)


jchalou:
Merged as part of https://github.com/graalvm/mx/commit/1e64cafd3b59f04c5e80d39fb8596b61253f6220.

eregon:
Thank you.
Would have been nice to preserve the author information (or `Co-authored-by:`) in https://github.com/graalvm/mx/commit/acecf78f86aaf6ea3aa43a796f0447e971598dd9 but not a big deal for such a small change.

